PR #7, opened 2019-12-02
3.0.0

BartoszLitwiniuk:


konraddysput:
One thing that could require our attention is travis CI tests. Its expected that our unit tests fail?

PR #13, opened 2020-12-18
Add more backtrace features to demo app

rqbacktrace:
Update demo app to include code samples for some Backtrace-Android features.

### DIFF
--- a/example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
+++ b/example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
@@ -3,6 +3,17 @@ package backtraceio.backtraceio;
 import android.content.Context;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.view.View;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.lang.reflect.Array;
+import java.nio.CharBuffer;
+import java.util.ArrayList;
+import java.util.Arrays;
 
 import backtraceio.library.BacktraceClient;
 import backtraceio.library.BacktraceCredentials;
@@ -11,8 +22,18 @@ import backtraceio.library.enums.database.RetryBehavior;
 import backtraceio.library.enums.database.RetryOrder;
 import backtraceio.library.models.BacktraceExceptionHandler;
 import backtraceio.library.models.database.BacktraceDatabaseSettings;
+import backtraceio.library.models.json.BacktraceReport;
 
 public class MainActivity extends AppCompatActivity {
+
+    private BacktraceClient backtraceClient;
+
+    private final int anrTimeout = 3000;
+
+    // Used to load the 'native-lib' library on application startup.
+    static {
+        System.loadLibrary("native-lib");
+    }
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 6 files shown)


example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -33,9 +43,37 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        database.setupNativeIntegration(backtraceClient, credentials);
+
+        // Enable ANR detection with a timeout of 3000 milliseconds
+        backtraceClient.enableAnr(3000);
+    }
+
+    public native void cppCrash();
+
+    public void handledException(View view) {
+        try {
```
konraddysput, line 61: From demo perspective - throwing this kind of exceeptions won't give any cool insights about exception for user. I would recommend:
* For example - make a method that will do a file operation on files that don't exist 
* For example - trying to access nth index of array with n-1 elements. 

In this situation user will see:
* inner methods in MainActivity to see how easy is to find a root cause of problems,
* will see real exceptions that happen in every managed android app,
* will see more details in web debugger.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -33,9 +43,37 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    public void unhandledException(View view) {
+        throw new NullPointerException("My null ptr exception");
+    }
+
+    public void nativeCrash(View view) {
+        cppCrash();
+    }
+
+    public void anr(View view) throws InterruptedException {
+        Thread.sleep(6000);
```
konraddysput, line 77: I think if user will change anr value for testing purposes and then fire "ANR" button, then he won't see anr in worst case. Instead I propose to expose ANR timeout to private const variable of a class and then in this method sleep for _anrTimeout + 1sec time.

konraddysput (changes requested):
Hey, I think we should do a few small design changes. Thank you for putting it together. it will definitely help.

konraddysput:
@rqbacktrace 

also Travis-ci fails for some reason beacuse of native integration. We must have it up to date and working correctly, otherwise we will need to do manual tests all the time :(

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
jasoncdavis0: I like this idea for more interesting exceptions well.

jasoncdavis0 (approved):
I'm not reviewing for code correctness, but this looks good to me.

BartoszLitwiniuk:
Travis is failing..

BartoszLitwiniuk:
besides Konrad's comments it is looks good to me

konraddysput:
> Travis is failing..

This is fixed in another branch - `TravisTest`. Anyway -let's validate if this change work after merging TravisTest with this branch.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
rqbacktrace: I had to increase the slack from your proposed 1sec to 2sec to force the ANR

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
rqbacktrace: I put some more interesting examples. 

The first is a handled exception for an equipment management interface, which is fairly typical for any game.

The second tries to read a save file (which is not there). But @konraddysput we know due to the existing inner exception issue we will not be able to catch the actual innermost exception for this case.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -33,9 +54,72 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    public native void cppCrash();
+
+    private ArrayList<String> equippedItems;
+
+    public ArrayList<String> getWarriorArmor()
+    {
+        return new ArrayList<String>(Arrays.asList("Tough Boots", "Strong Sword", "Sturdy Shield", "Magic Wand"));
+    }
+
+    int FindEquipmentIndex(ArrayList<String> armor, String equipment)
```
konraddysput, line 78: methods in java starts with lower case character - `findEquipmentIndex`

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -33,9 +54,72 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        backtraceClient.enableAnr(anrTimeout);
+    }
+
+    public native void cppCrash();
+
+    private ArrayList<String> equippedItems;
+
+    public ArrayList<String> getWarriorArmor()
+    {
+        return new ArrayList<String>(Arrays.asList("Tough Boots", "Strong Sword", "Sturdy Shield", "Magic Wand"));
```
konraddysput, line 75: It won't be better if you will use:
` List.of("Tough Boots", "Strong Sword", "Sturdy Shield", "Magic Wand");`?

Also - we don't need to pack it once again into `ArrayList` - `Arrays.asList("Tough Boots", "Strong Sword", "Sturdy Shield", "Magic Wand")` will generate List of strings.

konraddysput (changes requested):
Hey - small I think we should change two small things and we're ready to merge it to master. What do you think?

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
rqbacktrace: @konraddysput we are currently using Java 8 so we don't have `List.of` method available.

Also if I don't pack again into ArrayList, I have to cast it like this:
`return (ArrayList<String>) Arrays.asList("Tough Boots", "Strong Sword", "Sturdy Shield", "Magic Wand");`

I think because we are using older Java version we are stuck with this way for now.

PR #14, opened 2020-12-19
Allow user to inform Backtrace API if the incoming crashes are symbolicated with Proguard

rqbacktrace:
Allow user to inform Backtrace API if the incoming crashes are symbolicated with Proguard

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,8 @@ catch (e: Exception) {
 5. [Running sample application](#sample-app)
 6. [Using Backtrace library](#using-backtrace)
 7. [Working with NDK applications](#working_with_ndk)
-8. [Documentation](#documentation)
+8. [Working with Proguard](#working_with_proguard)
+9. [Documentation](#documentation)
 
 
 # Features Summary <a name="features-summary"></a>
@@ -407,6 +408,39 @@ More details about [extractNativeLibs](https://developer.android.com/guide/topic
 
 ## Uploading symbols to Backtrace
 For an NDK application, debugging symbols are not available to Backtrace by default. You will need to upload the application symbols for your native code to Backtrace. You can do this by uploading the native libraries themselves, which are usually found in the .apk bundle. [Click here to learn more about symbolification](https://support.backtrace.io/hc/en-us/articles/360040517071-Symbolication-Overview)
+
+# Working with Proguard <a name="working_with_proguard"></a>
+
+##### 1. Add the following to the `proguard_rules.pro` for your app
+These are needed since Proguard breaks some Backtrace libraries
+```
+-keep class com.google.gson.**.* { *; }
+-keep class backtraceio.library.**.* { *; }
+```
+##### 2. Enable Proguard mode in the [BacktraceClient](#documentation-BacktraceClient)
+```java
+backtraceClient.enableProguard();
+```
+
+##### 3. Create a UUID of your choice and set it as the value for the attribute `symbolication_id`, you will upload your Proguard mapping file with this same UUID later
+```java
+final UUID proguardMappingUUID = UUID.fromString("f6c3e8d4-8626-4051-94ec-53e6daccce25");
+final Map<String, Object> attributes = new HashMap<String, Object>() {{
+    put("symbolication_id", proguardMappingUUID.toString());
+}};
+```
+
+##### 4. Upload your Proguard mapping file with the UUID from the above step to Backtrace
+
+Currently we don't have a way to upload the Proguard mapping file from the UI. You will need to use a tool such as `curl` or Postman to upload the Proguard mapping file to Backtrace.
+To do so, please construct an HTTP POST request with the following parameters, and submit the mapping file as the request body: 
+`https://<Universe Name>.sp.backtrace.io:6098/post?format=proguard&token=<Symbol Upload Token>&universe=<Universe Name>&project=<Project Name>&symbolication_id=<symbolication_id from above>`
+
+##### 5. Start sending Proguard obfuscated crashes! 
+If the symbolication_id from the submitted crash matches a symbolication_id of a submitted Proguard mapping file, it will attempt to use that mapping file to deobfuscate the symbols from the submitted crash.
+
+#### Important Note for Windows users: 
+Please ensure your Proguard mapping file has Unix line endings before submitting to Backtrace!
 
 # Documentation  <a name="documentation"></a>
 
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -24,7 +24,6 @@ public class BacktraceClient extends BacktraceBase {
      */
     private BacktraceANRWatchdog anrWatchdog;
 
-
     /**
      * Initializing Backtrace client instance with BacktraceCredentials
      *
@@ -93,7 +92,6 @@ public class BacktraceClient extends BacktraceBase {
         super(context, credentials, database, attributes);
     }
 
-
     /**
      * Sending a message to Backtrace API
      *
@@ -152,7 +150,6 @@ public class BacktraceClient extends BacktraceBase {
             serverResponseEventListener) {
         super.send(report, serverResponseEventListener);
     }
-
 
     /**
      * Start monitoring if the main thread has been blocked
--- a/backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
+++ b/backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
@@ -27,6 +27,13 @@ public class BacktraceData {
     private static transient String LOG_TAG = BacktraceData.class.getSimpleName();
 
     /**
+     * If sending a Proguard obfuscated callstack, we need
+     * to set this field to "proguard" so the backend knows
+     */
+    @SerializedName("symbolication")
+    public String symbolication;
+
+    /**
      * 16 bytes of randomness in human readable UUID format
      * server will reject request if uuid is already found
      */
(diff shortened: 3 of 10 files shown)


backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
```diff
@@ -64,120 +68,140 @@
...
+            List<String> attachmentPaths,
+            Boolean isProguardEnabled)
+    {
         this.attributes = attributes == null ? new HashMap<String, Object>() {
         } : attributes;
         this.attachmentPaths = attachmentPaths == null ? new ArrayList<String>() : attachmentPaths;
         this.exception = exception;
         this.exceptionTypeReport = exception != null;
         this.diagnosticStack = new BacktraceStackTrace(exception).getStackFrames();
+	    this.isProguardEnabled = isProguardEnabled;
```
BartoszLitwiniuk, line 204: Wrong formatting - additional space

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -19,6 +19,11 @@
  */
 public class BacktraceClient extends BacktraceBase {
 
+    /**
+     * Is Proguard symbolication enabled? We have to inform the Backtrace API
+     */
+    private Boolean isProguardEnabled = false;
```
BartoszLitwiniuk, line 25: `isProguardEnabled ` can be null? If not it should be `boolean`

example-app/proguard-rules.pro
```diff
@@ -19,3 +19,7 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
```
BartoszLitwiniuk, line 22: it is worth to add it to README.md

BartoszLitwiniuk:
Maybe it is worth to add simple tests to confirm that everything is working properly

BartoszLitwiniuk (changes requested):
I left a few comments

backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
```diff
@@ -26,6 +26,13 @@
 
     private static transient String LOG_TAG = BacktraceData.class.getSimpleName();
 
+    /**
+     * 16 bytes of randomness in human readable UUID format
```
konraddysput, line 30: Why we add here a comment that this is a 16 bytes of randomness UUID if it's progruard?

konraddysput (changes requested):
We have a few problems with code formatting. Can you please apply default code formatting to all classes with your changes? 
Also, there is no information about this change in the changelog/readme file- should we also add this info there?

rqbacktrace:
Just updated the pull request, please have another look when you get the chance!

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
```diff
@@ -185,7 +185,11 @@ public BacktraceReport(
     }
 
     public BacktraceData toBacktraceData(Context context, Map<String, Object> clientAttributes) {
-        return new BacktraceData(context, this, clientAttributes);
+        return new BacktraceData(context, this, clientAttributes, false);
```
BartoszLitwiniuk, line 188: here you should call below method `toBacktraceData(context, this, clientAttributes, false) {`

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
```diff
@@ -185,7 +185,11 @@ public BacktraceReport(
     }
 
     public BacktraceData toBacktraceData(Context context, Map<String, Object> clientAttributes) {
```
BartoszLitwiniuk, line 187: there is one more thing which should be done, this method is used in the database (BacktraceDatabase.java) file, probably you will need to add `isProguardEnabled` parameter also there

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -150,9 +154,15 @@ public void send(BacktraceReport report) {
...
     public void send(BacktraceReport report, OnServerResponseEventListener
             serverResponseEventListener) {
-        super.send(report, serverResponseEventListener);
+        super.send(report, serverResponseEventListener, this.isProguardEnabled);
     }
 
+    /**
+     * Inform Backtrace API that we are using Proguard symbolication
+     */
+    public void enableProguard() {
```
konraddysput, line 163: Why we don't want to move `isProguardEnabled ` to `BacktraceBase` class? Advantages that I see:
* Even if someone will try to implement their own version of Backtrace integration by using `BacktraceBase` class, `isProguardEnabled` will still be necessary to add/implement.
* We will avoid passing isProguardEnabled flag every time to `BacktraceBase`
* More code in BacktraceClient class because of support  there for isProguardEnabled flag.

backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
```diff
@@ -110,17 +117,31 @@
...
+
+    /**
+     * Create instance of report data
+     *
+     * @param context           current application context
+     * @param report            current report
+     * @param clientAttributes  attributes which should be added to BacktraceData object
+     * @param isProguardEnabled is Proguard symbolication enabled?
+     */
+    public BacktraceData(Context context, BacktraceReport report, Map<String, Object>
```
konraddysput, line 137: Let's think about class design for a second. Is proguard flag necessary to create BacktraceData? I think in most cases - no. I think there is no reason to pass `isProguardEnabled` flag via the constructor. 

`symbolication` property also right now is public. 

Based on this info:
* `symbolication` property should be public? 
* I think we should use properties instead of constructor parameters to avoid unnecessary complexity and also to keep our code clean. Also, it's worth keeping in mind that if we will pass every type of configuration option via backtraceData constructor, our constructor will be huge.

My recommendation:
* set symbolication variable via setter not via the constructor.
* mark `symbolication` variable as internal and set it from `backtraceBase`

konraddysput (changes requested):
Hey, can we add few changes to class design/implementation?

Also, I think it's always worth to add description - what specific CR does.

README.md
```diff
@@ -408,6 +409,39 @@ More details about [extractNativeLibs](https://developer.android.com/guide/topic
...
+
+Currently we don't have a way to upload the Proguard mapping file from the UI. You will need to use a tool such as `curl` or Postman to upload the Proguard mapping file to Backtrace.
+To do so, please construct an HTTP POST request with the following parameters, and submit the mapping file as the request body: 
+`https://<Universe Name>.sp.backtrace.io:6098/post?format=proguard&token=<Symbol Upload Token>&universe=<Universe Name>&project=<Project Name>&symbolication_id=<symbolication_id from above>`
+
+##### 5. Start sending Proguard obfuscated crashes! 
+If the symbolication_id from the submitted crash matches a symbolication_id of a submitted Proguard mapping file, it will attempt to use that mapping file to deobfuscate the symbols from the submitted crash.
+
+#### Important Note for Windows users: 
+Please ensure your Proguard mapping file has Unix line endings before submitting to Backtrace!
```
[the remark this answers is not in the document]
rqbacktrace: Sure! Supporting Windows line endings will require a change to coronerd as we discussed before break but I'm happy to make that change as well! Should this be high-priority?

backtrace-library/src/androidTest/java/backtraceio/library/BacktraceClientSendTest.java
```diff
@@ -49,6 +55,7 @@ public void sendException() {
         RequestHandler rh = new RequestHandler() {
             @Override
             public BacktraceResult onRequest(BacktraceData data) {
+                assertNull(data.symbolication);
```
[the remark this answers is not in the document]
rqbacktrace: I understand. I will make a separate test class for Proguard

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
rqbacktrace: That makes sense, I will make this change

backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
rqbacktrace: I marked `symbolication` property as public since all other BacktraceData properties are public. 

Also I don't see `backtraceBase` and `backtraceClient` setting any BacktraceData fields currently, it seems we do pass all the options in from the constructor currently, and I didn't want to break that pattern. @BartoszLitwiniuk also suggested to add `isProguardEnabled` as a constructor parameter in BacktraceData. For now I don't know if we need to make this change to `symbolication`

konraddysput: @rqbacktrace I just though it will be better if we will set symbolication field from `backtracebase` or `backtraceClient`. We shouldn't pass backtraceClient/backtracebase instance to backtraceData.

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
rqbacktrace: I see this is much cleaner, thank you for this helpful feedback!

backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java
rqbacktrace: As per our conversation I think you are right @konraddysput , I have made this suggested change in the latest commit!

README.md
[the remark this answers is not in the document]
rqbacktrace: We have a separate ticket for Windows line ending support, no changes needed on the client side

konraddysput (approved):
Looks good for me :) Thanks!

konraddysput (approved):

BartoszLitwiniuk (approved):
LGTM, good job! 👍

PR #16, opened 2021-01-12
Expose DumpWithoutCrash to managed layer

rqbacktrace:
This commit exposes the crashpad function DumpWithoutCrash to the managed layer. This is useful if we want to debug something in the managed layer but at the same time we want to see the state of the native layer.

We don't need a unit test here since we are just exposing the crashpad function DumpWithoutCrash. I confirmed the behavior with manual testing.

### DIFF
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -120,6 +120,12 @@ extern "C" {
         *(volatile int *) 0 = 0;
     }
 
+    void DumpWithoutCrash() {
+        crashpad::NativeCPUContext context;
+        crashpad::CaptureContext(&context);
+        client->DumpWithoutCrash(&context);
+    }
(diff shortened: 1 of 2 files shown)


rqbacktrace:
Btw I think the CI failed since my DumpWithoutCrash branch was from an older version of master. It should not fail when it gets merged to master. It is obvious that no existing behavior is being changed here.

konraddysput (approved):
Looks cool to me! you committed empty CMakeLists.txt file in example-app dir - is this expected?

konraddysput:
@rqbacktrace this change will require to update our Crashpad libraries, right?

KrzaQ (approved):

rqbacktrace:
@konraddysput the cmakelists was a mistake, I will remove that file from this commit.

No change required to crashpad libraries, the crashpad library we are using has `DumpWithoutCrash` exposed already

PR #17, opened 2021-01-19
NDK - safe background thread operations

konraddysput:
Previously, Backtrace-android cached JNIEnv pointer on environment startup. While there is a small chance to initialize Backtrace-native integration in the background thread, there is a chance that someone would like to add an attribute or dump process from the background thread - for example when ANR/OOM occurred. 
 
Cached JNIEnv won't be available when we will call native code from managed background thread - which caused an exception. 

This diff allows to:
* Fetch dynamically JNIEnv when we will need to use it,
* format native code (was unformatted),
* extends `DumpWithoutCrash` method to include `error.message` that users define.

### DIFF
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -20,8 +20,8 @@ using namespace base;
 // supported JNI version
 static jint JNI_VERSION = JNI_VERSION_1_6;
 
-// JNI environment reference
-static JNIEnv *env;
+// Java VM
+static JavaVM *javaVm;
 
 // crashpad client
 static crashpad::CrashpadClient *client;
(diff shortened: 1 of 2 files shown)


KrzaQ (approved):

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -116,80 +158,119 @@ namespace /* anonymous */
...
-                                                          jobjectArray attributeValues) {
-        return Initialize(url, database_path, handler_path, attributeKeys, attributeValues);
-    }
+void DumpWithoutCrash(jstring message) {
+    crashpad::NativeCPUContext context;
+    crashpad::CaptureContext(&context);
 
+    // set dump message for single report
+    crashpad::SimpleStringDictionary *annotations = NULL;
+    bool did_attach = false;
```
rqbacktrace, line 172: Can we just call `AddAttribute` here as well to add the custom attribute `error.message` = `myMessage`? Why does it need to add the attribute in this custom way?

rqbacktrace (approved):
Looks good to me! I just have a question about how we should add the custom message to `DumpWithoutCrash`, but nothing which should block approval!

backtrace-library/src/main/cpp/backtrace-native.cpp
konraddysput: // user can't override error.message - exception message that Crashpad/crash-reporting tool
 // will set to tell user about error message. This code will set error.message only for single
 // report and after creating a dump, method will clean up this attribute.

We still need to later opearte on annotations object, so this is the reason why I prefered to have this code twice.

rqbacktrace: Ah okay I think I understand. This will set a one-time attribute "error.message" just for this specific dump called by `DumpWithoutCrash`. In contrast to the `AddAttribute` function which will add attributes for all subsequent dumps.

PR #18, opened 2021-01-27
Extended primitive type comparision

konraddysput:
Previously we used `.isPrimitiveType` method available in the `Object` class. This method doesn't return true, when a user pass - for example, primitive value 123. Type of value 123 is an Integer - not int. Because of that, we added this value to annotations, instead of to attributes.

This diff extends if statement and treat primitive object types like Integer, Long, Double like primitives.

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
+++ b/backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
@@ -14,6 +14,7 @@ import java.util.Locale;
 import java.util.Map;
 
 import backtraceio.library.BuildConfig;
+import backtraceio.library.common.TypeHelper;
 import backtraceio.library.enums.ScreenOrientation;
 
 /**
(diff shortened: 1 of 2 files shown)


backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
```diff
@@ -164,7 +165,7 @@ private void convertAttributes(BacktraceReport report, Map<String, Object> clien
                 continue;
             }
             Class type = value.getClass();
-            if (type.isPrimitive() || value instanceof String || type.isEnum()) {
+            if (TypeHelper.isPrimitiveOrPrimitiveWrapperOrString(type) || value instanceof String || type.isEnum()) {
```
rqbacktrace, line 168: Is `value instanceof String` still needed here? Looks like your new logic will do that too right?

rqbacktrace (approved):

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
konraddysput: Ah good catch. I will remove this check here. Thanks!

PR #20, opened 2021-03-05
Native report - include NDK main thread id

konraddysput:
**Why**

When the user creates a native report in a background thread, it's super useful for the user to also know what is the main thread id. 

**User story**
* as a Unity user when ANR occurs I would like to know what is main thread id to see what my main thread did when ANR occurred,
* as a Backtrace-Android user when I'm reporting application state from background thread I would like to know the main thread id,

This diff shows how to include the main thread id in default attributes in the initialization method.

### DIFF
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -31,6 +31,7 @@ static crashpad::CrashpadClient *client;
 // check if crashpad client is already initialized
 static std::atomic_bool initialized;
 static std::mutex attribute_synchronization;
+static std::string thread_id;
 
 JNIEXPORT jint JNI_OnLoad(JavaVM *jvm, void *reserved) {
     JNIEnv *env;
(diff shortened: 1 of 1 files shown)


rqbacktrace:
I understand the purpose of this addition.

I guess we should also tell the user somewhere to create their BacktraceClient from the main thread to ensure that it really will be the main thread ID that gets stored here.

Or maybe we should rename `thread.main` here to something like `thread.backtrace`? Just saying that this could potentially be confusing. Overall it looks right and I approve.

rqbacktrace (approved):

BartoszLitwiniuk (approved):
lgtm 👍

konraddysput:
@rqbacktrace good suggestion! I will check if we can get a thread id from JNI initialization method - then we will be sure that we will get a main thread id I guess.

rqbacktrace:
Based on discussion of your tests and research it seems this should work @konraddysput ! Looks good to me!

PR #21, opened 2021-03-06
Unreal engine improvements

rqbacktrace:
This contains 2 improvements we need for Unreal Engine:
- Include custom attributes in native crashes (that's why we fix the BacktraceAttributes constructor)
- Include custom file attachments in managed and native crashes

### DIFF
--- a/backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
+++ b/backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
@@ -47,7 +47,6 @@ public class SettingAttributesTest {
         clientAttributes = new HashMap<>();
 
         clientAttributes.put(customClientAttributeKey, customClientAttributeValue);
-
     }
 
     @Test
@@ -209,8 +208,10 @@ public class SettingAttributesTest {
 
     @Test
     public void ensureClientCustomAttributesPassedToNewBacktraceAttributes() {
-        BacktraceClient client = new BacktraceClient(context, backtraceCredentials, clientAttributes);
         BacktraceAttributes attributes = new BacktraceAttributes(context, null, clientAttributes);
-        assertEquals(clientAttributes.get("custom-client-attribute-id"), attributes.attributes.get("custom-client-attribute-id"));
+        assertNotNull(clientAttributes.get("custom-client-attribute-id"));
+        assertNotNull(attributes.attributes.get("custom-client-attribute-id"));
+        assertEquals(customClientAttributeValue, clientAttributes.get("custom-client-attribute-id"));
+        assertEquals(customClientAttributeValue, attributes.attributes.get("custom-client-attribute-id"));
     }
 }
--- a/backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
+++ b/backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
@@ -21,6 +21,7 @@ import backtraceio.library.models.BacktraceData;
 import backtraceio.library.models.BacktraceExceptionHandler;
 import backtraceio.library.models.BacktraceResult;
 import backtraceio.library.models.database.BacktraceDatabaseSettings;
+import backtraceio.library.models.json.BacktraceAttributes;
 import backtraceio.library.models.types.BacktraceResultStatus;
 
 import static org.junit.Assert.assertEquals;
@@ -204,5 +205,12 @@ public class SettingAttributesTest {
         } catch (Exception exception) {
             Assert.fail(exception.getMessage());
         }
+    }
+
+    @Test
+    public void ensureClientCustomAttributesPassedToNewBacktraceAttributes() {
+        BacktraceClient client = new BacktraceClient(context, backtraceCredentials, clientAttributes);
+        BacktraceAttributes attributes = new BacktraceAttributes(context, null, clientAttributes);
+        assertEquals(clientAttributes.get("custom-client-attribute-id"), attributes.attributes.get("custom-client-attribute-id"));
     }
 }
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -2,6 +2,7 @@ package backtraceio.library;
 
 import android.content.Context;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -131,7 +132,7 @@ public class BacktraceClient extends BacktraceBase {
      */
     public BacktraceClient(Context context, BacktraceCredentials credentials,
                            Database database) {
-        this(context, credentials, database, (Map<String, Object>) null);
+        this(context, credentials, database, new HashMap<String, Object>());
     }
 
     /**
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -2,6 +2,7 @@ package backtraceio.library;
 
 import android.content.Context;
 
+import java.util.List;
 import java.util.Map;
 
 import backtraceio.library.base.BacktraceBase;
@@ -29,7 +30,7 @@ public class BacktraceClient extends BacktraceBase {
      * @param credentials credentials to Backtrace API server
      */
     public BacktraceClient(Context context, BacktraceCredentials credentials) {
-        this(context, credentials, (BacktraceDatabase) null, null);
+        this(context, credentials, (BacktraceDatabase) null);
     }
 
     /**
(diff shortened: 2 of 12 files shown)


backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -371,6 +458,7 @@ public void send(BacktraceReport report, final OnServerResponseEventListener cal
         if (breadcrumbs != null) {
             breadcrumbs.processReportBreadcrumbs(report);
         }
+        processReportAttachments(report);
```
BartoszLitwiniuk, line 461: maybe better name can be 'addReportAttachments'?

BartoszLitwiniuk (approved):
Good job 👍

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
```diff
@@ -48,9 +48,15 @@ public BacktraceAttributes(Context context, BacktraceReport report, Map<String,
             clientAttributes) {
         this.context = context;
         if (report != null) {
-            this.convertAttributes(report, clientAttributes);
+            this.convertReportAttributes(report);
             this.setExceptionAttributes(report);
         }
+        if (clientAttributes != null) {
```
konraddysput, line 54: Why do we need this change in this diff? What is the purpose of this change?

backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
```diff
@@ -205,4 +206,11 @@ public void uncaughtException(Thread t, Throwable e) {
             Assert.fail(exception.getMessage());
         }
     }
+
+    @Test
+    public void ensureClientCustomAttributesPassedToNewBacktraceAttributes() {
+        BacktraceClient client = new BacktraceClient(context, backtraceCredentials, clientAttributes);
+        BacktraceAttributes attributes = new BacktraceAttributes(context, null, clientAttributes);
+        assertEquals(clientAttributes.get("custom-client-attribute-id"), attributes.attributes.get("custom-client-attribute-id"));
```
konraddysput, line 214: Why you're testing if you can get an attribute instead of test also the value of attribute? This method I believe, can return null which might be unexpected behaviour.

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -75,7 +131,20 @@ public BacktraceClient(Context context, BacktraceCredentials credentials,
      */
     public BacktraceClient(Context context, BacktraceCredentials credentials,
                            Database database) {
-        this(context, credentials, database, null);
+        this(context, credentials, database, (Map<String, Object>) null);
```
konraddysput, line 134: Does it make sense to make a cast that will return null instead of passing an empty dictionary to avoid confusion? Later you need to check anyway if the dictionary exists.

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -145,7 +147,13 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
         String[] values = crashpadAttributes.attributes.values().toArray(new String[0]);
 
         // Paths to Crashpad attachments
-        String[] attachmentPaths = new String[] {this.breadcrumbs.getBreadcrumbLogPath()};
+        List<String> attachmentPaths = new ArrayList<String>();
```
konraddysput, line 150: Do we need to create a list? We cast it anyway to array later, so we will avoid unnecessary cast later and make code simpler with just array. We also know array size because we know how many attachments we store + additional breadcrumbs file.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
konraddysput: +1

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -152,6 +170,17 @@ public void sendReport(View view) {
         backtraceClient.send(report);
     }
 
+    public void writeMyCustomFile(String filePath) {
```
konraddysput, line 173: This method should be private.

konraddysput (commented):
Added few sugestions to your code to make maybe code simpler/prettier - please let me know what you think about them. I will be really happy if we can adjust few places.


Also - did you test backtraceDatabaseReport? I don't see any test with database records and new attachment system. Does backtraceDatabaseRecord store attachment path and when database try to resend a report, database report will have attachments?

Otherwise code looks fine and won't break. Thanks for this change!

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
rqbacktrace: The purpose of this change is that we don't need to have a valid `report` to take the `clientAttributes`. If we have valid `clientAttributes`, we should include those in the new `BacktraceAttributes` even if we don't have a `report`. 

This was preventing `BacktraceAttributes` from being constructed properly for Crashpad. Because we were calling `BacktraceAttributes` constructor with a null `report`, so the `clientAttributes` were not getting passed to Crashpad. Now with this change, we do get the `clientAttributes` from Crashpad, even though we call `BacktraceAttributes` constructor with a null `report`.

backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
rqbacktrace: This tests that the value of `clientAttributes["custom-client-attribute-id"] == attributes.attributes["custom-client-attribute-id"]`. I will add null checks too to make it more robust.

Dautery91 (approved):
FYI I just took a quick skim here, pushing through so I'm not blocking you.  Looked reasonable to me, as did Konrad's feedback.

backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
konraddysput: but also can we be sure that `clientAttributes["custom-client-attribute-id"]` store value that it suppose to store?

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
konraddysput: oh now I understand! thanks for explanation Rameez!

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
rqbacktrace: If we use a Java String Array here, we get some pretty messy hard to maintain code:
`        String[] breadcrumbAttachmentPath = {this.breadcrumbs.getBreadcrumbLogPath()};
        String[] clientAttachmentPaths = (String[]) client.attachments.toArray();
        String[] attachmentPaths = new String[breadcrumbAttachmentPath.length + clientAttachmentPaths.length];
        System.arraycopy(breadcrumbAttachmentPath, 0, attachmentPaths, 0, breadcrumbAttachmentPath.length);
        System.arraycopy(clientAttachmentPaths, 0, attachmentPaths, breadcrumbAttachmentPath.length, clientAttachmentPaths.length);`
        
In general we prefer a List<String> in Java wherever possible, to avoid messy code like the above, see: https://stackoverflow.com/questions/2843366/how-to-add-new-elements-to-an-array

But we prefer a String Array for native interface, so best to work with a List<String> while we're in Java and convert to String array when we want to go to native code.

backtrace-library/src/androidTest/java/backtraceio/library/SettingAttributesTest.java
rqbacktrace: Will do, thank you!

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
rqbacktrace: Will do, thank you!

rqbacktrace:
Hey thank you for the point about testing database record @konraddysput ! I actually found a bug there by making this test! Will be pushing the fix shortly

rqbacktrace:
@BartoszLitwiniuk @konraddysput [please take a look at the latest commit in this PR](https://github.com/backtrace-labs/backtrace-android/pull/21/commits/e1fc07c45e3903380501a270308f5770e8328093). 

In this commit I fix a bug with `BacktraceData.getAttachments`. 
When `BacktraceData` is deserialized from a `BacktraceDatabaseRecord`, we lose the `Context`. But `BacktraceData.getAttachments` relies on a valid context, otherwise it breaks. So now I require `BacktraceData.getBacktraceData` to take a `Context`, so that when we deserialize the `BacktraceDatabaseRecord`, we can restore the `Context`. GSON serialization ignores the `Context`, otherwise this wouldn't be necessary anyway.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -400,4 +487,12 @@ public void onEvent(BacktraceResult backtraceResult) {
             }
         };
     }
+
+    private void addReportAttachments(BacktraceReport report) {
```
BartoszLitwiniuk, line 491: Am I correct that we will add this attachment to every report? Is this the expected behavior?

rqbacktrace: Yes and yes

backtrace-library/src/main/java/backtraceio/library/common/BacktraceSerializeHelper.java
```diff
@@ -33,6 +33,7 @@ public static String toJson(Object object) {
     }
 
     public static <T> T fromJson(String json, Class<T> type) {
-        return new Gson().fromJson(json, type);
+        Gson gson = new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_DASHES).create();
```
BartoszLitwiniuk, line 36: maybe it is worth  to extract this line to private method in this file and use it in both places to make sure that we will not have any issues like this in the future

BartoszLitwiniuk (approved):
Good job!

backtrace-library/src/main/java/backtraceio/library/common/BacktraceSerializeHelper.java
rqbacktrace: Done, thank you for the excellent advice!

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
rqbacktrace: Thanks for sharing the example from Slack @konraddysput , we are now using String[] here!

PR #23, opened 2021-03-15
Set main thread as a faulting thread in DumpWithoutCrash API

konraddysput:
# Why

When ANR occurred, we need to switch the main thread to the background thread to dump the process. In this situation, Crashpad will report the background thread as a faulting thread. This change will let us set the main thread as a faulting thread when ANR occurred OR when we know we should display the main thread as a faulting thread when we dumping a process from the background thread.

### DIFF
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -195,7 +195,7 @@ void Crash() {
     *(volatile int *) 0 = 0;
 }
 
-void DumpWithoutCrash(jstring message) {
+void DumpWithoutCrash(jstring message, jboolean set_main_thread_as_faulting_thread) {
     crashpad::NativeCPUContext context;
     crashpad::CaptureContext(&context);
 
@@ -203,7 +203,7 @@ void DumpWithoutCrash(jstring message) {
     crashpad::SimpleStringDictionary *annotations = NULL;
     bool did_attach = false;
 
-    if (message != NULL) {
+    if (message != NULL || set_main_thread_as_faulting_thread == true) {
         JNIEnv *env = GetJniEnv();
         if (env == nullptr) {
             __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Cannot initialize JNIEnv");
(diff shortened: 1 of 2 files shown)


backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -216,14 +216,18 @@ void DumpWithoutCrash(jstring message) {
...
-        // will set to tell user about error message. This code will set error.message only for single
-        // report and after creating a dump, method will clean up this attribute.
-        jboolean isCopy;
-        const char *rawMessage = env->GetStringUTFChars(message, &isCopy);
-        annotations->SetKeyValue("error.message", rawMessage);
-        env->ReleaseStringUTFChars(message, rawMessage);
+        if(set_main_thread_as_faulting_thread == true) {
+            annotations->SetKeyValue("_mod_faulting_tid", thread_id);
+        }
+        if(message != NULL) {
```
rqbacktrace, line 222: Why do we repeat this `message != NULL` check? Maybe we can fix this now or later

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -195,15 +195,15 @@ void Crash() {
...
+void DumpWithoutCrash(jstring message, jboolean set_main_thread_as_faulting_thread) {
     crashpad::NativeCPUContext context;
     crashpad::CaptureContext(&context);
 
     // set dump message for single report
     crashpad::SimpleStringDictionary *annotations = NULL;
     bool did_attach = false;
 
-    if (message != NULL) {
+    if (message != NULL || set_main_thread_as_faulting_thread == true) {
```
rqbacktrace, line 206: These checks and behaviors seem like they are independent. Maybe we can break up this if statement?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -306,8 +310,16 @@ Java_backtraceio_library_BacktraceDatabase_addAttribute(JNIEnv *env, jobject thi
 }
 
 JNIEXPORT void JNICALL
-Java_backtraceio_library_base_BacktraceBase_dumpWithoutCrash(JNIEnv *env, jobject thiz,
-                                                             jstring message) {
-    DumpWithoutCrash(message);
+Java_backtraceio_library_base_BacktraceBase_dumpWithoutCrash__Ljava_lang_String_2(JNIEnv *env,
```
rqbacktrace, line 313: Why did this function name change even though arguments are the same? Maybe we don't need to care about this

rqbacktrace (approved):
Looks good to me! Just had one suggestion about breaking up an if statement in `DumpWithoutCrash`

backtrace-library/src/main/cpp/backtrace-native.cpp
konraddysput: I did that for one purpose  - if user didn't define message and user don't want to change faulting thread to main thread - there is no need to read Crashpad attributes and set them.

backtrace-library/src/main/cpp/backtrace-native.cpp
konraddysput: Because we extended the current function to prevent API break.

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: Ah okay that makes sense

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: Ah okay sounds good. Thank you!

PR #24, opened 2021-05-07
Battery status

ianrice07:
Modifies how the Backtrace-Android library gets the battery status.  Uses a subscriber instead of querying the system.

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -196,12 +196,16 @@ public class BacktraceClient extends BacktraceBase {
         batteryStateHelper = BatteryStateHelper.getInstance(context, database);
     }
 
-    public void setBatteryMonitoring(boolean enable) {
+    public void enableBatteryMonitoring(boolean enable) {
         if (enable) {
             batteryStateHelper.enable();
         } else {
             batteryStateHelper.disable();
         }
+    }
+
+    public boolean isMonitoringBattery() {
+        return batteryStateHelper.isEnabled();
     }
 
     /**
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -193,25 +193,14 @@ public class BacktraceClient extends BacktraceBase {
                            Database database, Map<String, Object> attributes,
                            List<String> attachments) {
         super(context, credentials, database, attributes, attachments);
+        batteryStateHelper = BatteryStateHelper.getInstance(context, database);
     }
 
-    public void setBatteryMonitoring(boolean state) {
-        monitoringBattery = state;
-        if (state) {
-                if (batteryStateHelper == null)
-                    batteryStateHelper = new BatteryStateHelper(context, database);
-                batteryStateHelper.enable();
-                if (ifilter == null) {
-                    ifilter = new IntentFilter();
-                    ifilter.addAction(Intent.ACTION_BATTERY_CHANGED);
-                }
-                if (batteryStateHelper != null && ifilter != null) {
-                    context.registerReceiver(batteryStateHelper, ifilter);
-                }
+    public void setBatteryMonitoring(boolean enable) {
+        if (enable) {
+            batteryStateHelper.enable();
         } else {
-            if (batteryStateHelper != null)
-                batteryStateHelper.disable();
-                context.unregisterReceiver(batteryStateHelper);
+            batteryStateHelper.disable();
         }
     }
 
@@ -328,20 +317,6 @@ public class BacktraceClient extends BacktraceBase {
     public void disableAnr() {
         if (this.anrWatchdog != null && !this.anrWatchdog.isInterrupted()) {
             this.anrWatchdog.stopMonitoringAnr();
-        }
-    }
-
-    public void onPause() {
-        if(monitoringBattery) {
-            if (batteryStateHelper != null)
-                context.unregisterReceiver(batteryStateHelper);
-        }
-    }
-
-    public void onResume() {
-        if(monitoringBattery) {
-            if (batteryStateHelper != null && ifilter != null)
-                context.registerReceiver(batteryStateHelper, ifilter);
         }
     }
 }
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -1,6 +1,8 @@
 package backtraceio.library;
 
 import android.content.Context;
+import android.content.IntentFilter;
+import android.content.Intent;
 
 import java.util.HashMap;
 import java.util.List;
@@ -13,6 +15,7 @@ import backtraceio.library.models.database.BacktraceDatabaseSettings;
 import backtraceio.library.models.json.BacktraceReport;
 import backtraceio.library.watchdog.BacktraceANRWatchdog;
 import backtraceio.library.watchdog.OnApplicationNotRespondingEvent;
+import backtraceio.library.common.BatteryStateHelper;
 
 /**
  * Backtrace Java Android Client
@@ -23,6 +26,21 @@ public class BacktraceClient extends BacktraceBase {
      * Backtrace ANR watchdog instance
      */
     private BacktraceANRWatchdog anrWatchdog;
+
+    /**
+     * Backtrace Battery Helper
+     */
+     private BatteryStateHelper batteryStateHelper;
+
+    /**
+     * Filter for the batteryStateHelper
+     */
+    private IntentFilter ifilter;
+
+    /**
+     * Battery monitoring enabled
+     */
+    private boolean monitoringBattery = false;
 
     /**
      * Initializing Backtrace client instance with BacktraceCredentials
@@ -177,6 +195,26 @@ public class BacktraceClient extends BacktraceBase {
         super(context, credentials, database, attributes, attachments);
     }
 
+    public void setBatteryMonitoring(boolean state) {
+        monitoringBattery = state;
+        if (state) {
+                if (batteryStateHelper == null)
+                    batteryStateHelper = new BatteryStateHelper(context, database);
+                batteryStateHelper.enable();
+                if (ifilter == null) {
+                    ifilter = new IntentFilter();
+                    ifilter.addAction(Intent.ACTION_BATTERY_CHANGED);
+                }
+                if (batteryStateHelper != null && ifilter != null) {
+                    context.registerReceiver(batteryStateHelper, ifilter);
+                }
+        } else {
+            if (batteryStateHelper != null)
+                batteryStateHelper.disable();
+                context.unregisterReceiver(batteryStateHelper);
+        }
+    }
+
     /**
      * Sending a message to Backtrace API
      *
@@ -290,6 +328,20 @@ public class BacktraceClient extends BacktraceBase {
     public void disableAnr() {
         if (this.anrWatchdog != null && !this.anrWatchdog.isInterrupted()) {
             this.anrWatchdog.stopMonitoringAnr();
+        }
+    }
+
+    public void onPause() {
+        if(monitoringBattery) {
+            if (batteryStateHelper != null)
+                context.unregisterReceiver(batteryStateHelper);
+        }
+    }
+
+    public void onResume() {
+        if(monitoringBattery) {
+            if (batteryStateHelper != null && ifilter != null)
+                context.registerReceiver(batteryStateHelper, ifilter);
         }
     }
 }
--- a/backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
+++ b/backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
@@ -1,0 +1,198 @@
+package backtraceio.library.common;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.os.BatteryManager;
+
+
+import java.util.HashMap;
+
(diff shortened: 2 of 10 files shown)


backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -24,6 +27,21 @@
...
 
+    /**
+     * Backtrace Battery Helper
+     */
+     private BatteryStateHelper batteryStateHelper;
+
+    /**
+     * Filter for the batteryStateHelper
+     */
+    private IntentFilter ifilter;
```
rqbacktrace, line 38: `ifilter` should probably be an internal field of `batteryStateHelper`

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -292,4 +330,18 @@ public void disableAnr() {
             this.anrWatchdog.stopMonitoringAnr();
         }
     }
+
+    public void onPause() {
+        if(monitoringBattery) {
```
rqbacktrace, line 335: Most of the logic in this function should probably be moved to BatteryStateHelper

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -292,4 +330,18 @@ public void disableAnr() {
...
+            if (batteryStateHelper != null)
+                context.unregisterReceiver(batteryStateHelper);
+        }
+    }
+
+    public void onResume() {
+        if(monitoringBattery) {
```
rqbacktrace, line 342: Most of the logic in this function should probably be moved to BatteryStateHelper

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -24,6 +27,21 @@
      */
     private BacktraceANRWatchdog anrWatchdog;
 
+    /**
+     * Backtrace Battery Helper
+     */
+     private BatteryStateHelper batteryStateHelper;
```
rqbacktrace, line 33: Maybe the interface should be called `BatteryStateHelper` and the implementation called `BacktraceBatteryStateHelper`. This is just a nit though.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -184,4 +190,16 @@ private void writeMyCustomFile(String filePath) {
     public void exit(View view) {
         System.exit(0);
     }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        backtraceClient.onResume();
```
rqbacktrace, line 197: Requiring the user to do this will make `BacktraceClient` much more complicated to use for the customer. There's probably a way for `BacktraceClient` to capture these signals without the user having to do something, using [ActivityLifecycleCallbacks ](https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks) for instance?

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -43,6 +45,7 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        BacktraceLogger.setLevel(LogLevel.ERROR);
```
rqbacktrace, line 48: I don't think we need this in the demo activity

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -73,6 +76,7 @@ protected void onCreate(Bundle savedInstanceState) {
         BacktraceDatabase database = new BacktraceDatabase(context, settings);
         backtraceClient = new BacktraceClient(context, credentials, database, attributes, attachments);
 
+
```
rqbacktrace, line 79: Extra newline

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -57,14 +57,13 @@ public DeviceAttributesHelper(Context context) {
         result.put("device.gps.enabled", getGpsStatus().toString());
         result.put("device.bluetooth_status", isBluetoothEnabled().toString());
         result.put("device.cpu.temperature", String.valueOf(getCpuTemperature()));
-        result.put("device.is_power_saving_mode", String.valueOf(isPowerSavingMode()));
+        //result.put("device.is_power_saving_mode", String.valueOf(isPowerSavingMode()));
```
rqbacktrace, line 60: Why do we remove this?

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -187,20 +186,23 @@ private WifiStatus getWifiStatus() {
      * @return is power saving mode activated
      */
     // TODO: replace bool to enum
+    /*
```
rqbacktrace, line 189: Why do we remove this? I also don't see this power saving mode check re-implemented anywhere else?

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -187,20 +186,23 @@ private WifiStatus getWifiStatus() {
...
         if (Build.VERSION.SDK_INT < 21) {
             return false;
         }
         PowerManager powerManager = (PowerManager) this.context.getSystemService(Context
                 .POWER_SERVICE);
         return powerManager.isPowerSaveMode();
-    }
+    }*/
 
+    // Try removing this since it's not the correct way to do this
```
rqbacktrace, line 199: If we trust the new method from `BatteryStateHelper` to do this now, we can just remove this, no need to comment it out

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -187,20 +186,23 @@ private WifiStatus getWifiStatus() {
...
     /**
      * Get a battery level in float value (from 0.0 to 1.0) or -1 if error occurs
      *
      * @return battery level from 0.0 to 1.0 or -1 if error occurs
      */
+     /*
     private float getBatteryLevel() {
```
rqbacktrace, line 206: If we trust the new method from `BatteryStateHelper` to do this now, we can just remove this, no need to comment it out. But only if we're okay with users no longer having this information by default and needing to explicitly enable `BatteryStateHelper` to get this information now.

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,198 @@
...
+
+import java.util.HashMap;
+
+import backtraceio.library.BacktraceDatabase;
+import backtraceio.library.interfaces.Database;
+
+public class BatteryStateHelper extends BroadcastReceiver {
+    private static Context context;
+    private static BacktraceDatabase database;
+    private static boolean trackBattery;
```
rqbacktrace, line 18: You can rename this to something like `enabled`

rqbacktrace, line 18: Consider renaming `trackBattery` to something like `enabled`

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    public void disable() {
+        trackBattery = false;
+    }
+
+    public void enable() {
+        trackBattery = true;
+        IntentFilter ifilter = new IntentFilter(Intent.ACTION_BATTERY_CHANGED);
```
rqbacktrace, line 60: Why do we need to register an intent filter for the same action twice? Maybe we can remove this redundancy by doing all the `context.registerReceiver` calls from inside here when we move them out of `BacktraceClient`.

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,198 @@
...
+        Intent batteryStatus = context.registerReceiver(null, ifilter);
```
rqbacktrace, line 61: Why do we need to register an intent filter for the same action twice? Maybe we can remove this redundancy by doing all the `context.registerReceiver` calls from inside here when we move them out of `BacktraceClient`.

rqbacktrace (changes requested):
This is really cool, thank you @ianrice07! I think we just need to make sure to avoid polluting the main `BacktraceClient` class too much (we will also unify a lot of duplicate logic this way), and we should avoid making the API much heavier for users by forcing the user to manually register `BacktraceClient` activity lifecycle callbacks.

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -24,6 +27,21 @@
...
+
+    /**
+     * Filter for the batteryStateHelper
+     */
+    private IntentFilter ifilter;
+
+    /**
+     * Battery monitoring enabled
+     */
+    private boolean monitoringBattery = false;
```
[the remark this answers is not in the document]
ianrice07: Moved to be internal to BatteryStateHelper completely.

gradle.properties
```diff
@@ -30,4 +30,6 @@ POM_DEVELOPER_NAME=Bartosz Litwiniuk
 POM_DEVELOPER_ID2=KonradDysput
 POM_DEVELOPER_NAME2=Konrad Dysput
 POM_DEVELOPER_ID3=rqbacktrace
-POM_DEVELOPER_NAME3=Rameez Qurashi
\ No newline at end of file
+POM_DEVELOPER_NAME3=Rameez Qurashi
+android.useAndroidX=true
```
rqbacktrace, line 34: Android X

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
rqbacktrace: I think we still have an extra newline here

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,244 @@
...
+            getBatteryChargingStatus(batteryStatus);
+            getBatteryLevel(batteryStatus);
+            getBatteryTechnology(batteryStatus);
+            getBatteryTemp(batteryStatus);
+            if (database != null) {
+                updateDatabase();
+            }
+            context.registerReceiver(batteryStateHelper, intentFilter);
+        } else {
+            BacktraceLogger.d(TAG,"Context not set.  Run getInstance(context, database) first");
```
rqbacktrace, line 100: This should be an error

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,244 @@
...
+            context.unregisterReceiver(batteryStateHelper);
+            ProcessLifecycleOwner.get().getLifecycle().removeObserver(batteryStateHelper);
+        }
+    }
+
+    public void enable() {
+        if (getInstance(null, null) != null) {
+            enabled = true;
+            ProcessLifecycleOwner.get().getLifecycle().addObserver(batteryStateHelper);
+            Intent batteryStatus = context.registerReceiver(null, intentFilter);
```
rqbacktrace, line 88: Do we need to make two `context.registerReceiver` calls in this function? Can we remove this line?

backtrace-library/src/main/java/backtraceio/library/common/BatteryStateHelper.java
```diff
@@ -0,0 +1,244 @@
...
+                BatteryStateHelper.database = (BacktraceDatabase)database;
+            }
+        }
+        if (intentFilter == null) {
+            intentFilter = new IntentFilter();
+            intentFilter.addAction(Intent.ACTION_BATTERY_CHANGED);
+        }
+    }
+
+    public static BatteryStateHelper getInstance(Context context, Database database) {
```
rqbacktrace, line 43: Nit: This doesn't need to be a singleton. It would make your `enable` and `disable` functions cleaner, avoid having to make your functions `static`, and you can just use the keyword `this` when you want to get the current class instance. There should be a good reason for making a class a Singleton.

backtrace-library/src/main/java/backtraceio/library/common/PermissionHelper.java
```diff
@@ -3,7 +3,9 @@
 import android.Manifest;
 import android.content.Context;
 import android.content.pm.PackageManager;
-import android.support.v4.content.ContextCompat;
+
+import androidx.core.content.ContextCompat;
```
rqbacktrace, line 7: AndroidX

rqbacktrace (changes requested):
Thank you @ianrice07, this is much more self-contained and easier to use now! I think we should avoid using AndroidX though, as this will break backwards compatibility with our users who may not be using AndroidX yet. It also seems AndroidX requires API 28+, but we currently support API 21+. https://developer.android.com/jetpack/androidx#using_androidx_libraries_in_your_project

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
rqbacktrace: Do we want to set the `LogLevel.ERROR` here? I think for first-time users of our project it could be useful to see the most verbose Backtrace logs by default.

rqbacktrace:
Closing this for now to avoid polluting PR page but we will reopen this once we migrate to AndroidX

rqbacktrace:
Please do not delete this branch as we will revisit this later

PR #25, opened 2021-05-08
Fix database timer logging

BartoszLitwiniuk:


.travis.yml
```diff
@@ -41,8 +41,14 @@ install:
   - echo y | sdkmanager 'ndk;20.0.5594570'
   - echo y | sdkmanager "cmake;3.10.2.4988404"
 before_script:
+  - sudo apt-get update
+  - sudo apt-get install python3
+  - sudo apt-get install python3-pip
   - chmod +x gradlew
-  - pip install --user codecov #Install codecov
+  - python --version
```
konraddysput, line 48: do we need to print python versions here?

konraddysput (approved):

.travis.yml
BartoszLitwiniuk: Let's keep it because we will be quickly able to compare it if Travis will fail in the future. Then it can be first check if python version changed.

PR #26, opened 2021-06-21
Feature/client side unwinding

rqbacktrace:
This change allows client side unwinding. If client side unwinding is enabled, then when there is a native/crashpad crash, the call stack will be resolved by the device (client) before submission to the Backtrace server. This allows better resolution of call stacks 
when symbols for native libraries are not available, e.g for syscalls or opaque libraries.

### DIFF
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -118,8 +118,7 @@ namespace /* anonymous */
         bun_handle handle;
 
         bool bun_initialized = bun_handle_init(&handle, BUN_BACKEND_LIBUNWINDSTACK);
-        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "%s",
-                            "Bun initialized? %d", bun_initialized);
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Bun initialized? %d", bun_initialized);
         if (!bun_initialized) {
             return -1;
         }
@@ -127,8 +126,7 @@ namespace /* anonymous */
         bun_buffer buf = { buffer_child, BUFFER_SIZE };
 
         auto written = bun_unwind_remote(&handle, &buf, tid);
-        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "%s",
-                            "Bun bytes written %d", written);
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Bun bytes written %d", written);
         (void) written;
 
         return -1;
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -19,6 +19,7 @@
 
 #include "bun/bun.h"
 #include "bun/stream.h"
+#include "libbun/src/bun_internal.h"
 
 #include <sys/syscall.h>
 
@@ -43,6 +44,8 @@ struct bun_handle handle;
 // bun buffer
 char buf[65536];
 
+struct bun_buffer bun_buf;
+
 /*
  * Signal handler executed by CrashpadClient::SetFirstChanceExceptionHandler.
  */
@@ -52,7 +55,7 @@ static bool bun_sighandler(int signum, siginfo_t *info, ucontext_t *context)
     (void) info;
     (void) context;
 
-    bun_unwind(&handle, buf, sizeof(buf));
+    bun_unwind(&handle, &bun_buf);
 
     return false;
 }
@@ -358,6 +361,10 @@ bool EnableClientSideUnwinding() {
     if (!bun_handle_init(&handle, BUN_BACKEND_LIBUNWINDSTACK)) {
         __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
                             "Could not initialize bun_handle");
+    }
+    if (!bun_buffer_init(&bun_buf, buf, sizeof(buf))) {
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
+                            "Could not initialize bun_buffer");
     }
     crashpad::CrashpadInfo::GetCrashpadInfo()
             ->AddUserDataMinidumpStream(BUN_STREAM_ID, buf, sizeof(buf));
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -230,16 +230,14 @@ namespace /* anonymous */
 
         thread_local bool flag;
 
-        if (flag == false) {
-            flag = true;
-            bcd_emit(&bcd, "1");
-
-            crashpad::CrashpadClient::DumpWithoutCrash(context);
-
-            return true;
-        } else {
+        if (flag == true) {
             return false;
         }
+
+        flag = true;
+        bcd_emit(&bcd, "1");
+
+        crashpad::CrashpadClient::DumpWithoutCrash(context);
 
         return true;
     }
--- a/backtrace-library/src/main/cpp/backtrace-native.cpp
+++ b/backtrace-library/src/main/cpp/backtrace-native.cpp
@@ -119,6 +119,8 @@ namespace /* anonymous */
 
         bool bun_initialized = bun_handle_init(&handle, BUN_BACKEND_LIBUNWINDSTACK);
         __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Bun initialized? %d", bun_initialized);
+        __android_log_print(ANDROID_LOG_DEBUG, "Backtrace-Android", "request handler tid %ju, my pid %ju", (uintmax_t ) tid, (uintmax_t ) getpid());
+
         if (!bun_initialized) {
             return -1;
         }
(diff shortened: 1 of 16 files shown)


backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -321,4 +348,26 @@ Java_backtraceio_library_base_BacktraceBase_dumpWithoutCrash__Ljava_lang_String_
                                                                                    jboolean set_main_thread_as_faulting_thread) {
     DumpWithoutCrash(message, set_main_thread_as_faulting_thread);
 }
+
+bool EnableClientSideUnwinding() {
```
rqbacktrace, line 352: @konraddysput here is the native function you can use to enable client-side unwinding from backtrace-unity

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -429,6 +428,25 @@ public boolean addBreadcrumb(String message, Map<String, Object> attributes, Bac
         return database.getBreadcrumbs().addBreadcrumb(message, attributes, type, level);
     }
 
+    /**
+     * Enable client-side callstack resolution
+     * @param path              path to use for the socket file if DUMPWITHOUTCRASH unwinding is used, can be null
+     * @return true on success, otherwise returns false
+     */
+    public boolean enableClientSideUnwinding(String path) {
```
konraddysput, line 436: Why do we need to pass a path to client side unwinding specific file. Why we can't maintain it inside backtrace-database directory or inside crashpad dir? In both examples - please remember we need to make sure directory exist. If we want to let users use it - we can also make sure they can't shoot into their foot via our API.

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -48,6 +91,167 @@ JNIEXPORT jint JNI_OnLoad(JavaVM *jvm, void *reserved) {
...
+
+    bool RemoteUnwindingHandlerDumpWithoutCrash(int signum, siginfo_t *info, ucontext_t *context)
+    {
+        (void) signum;
+        (void) info;
+        (void) context;
+
+        thread_local bool flag;
+
+        if (flag == false) {
```
konraddysput, line 222: this condition could be simplify 
```
if (flag = true) {
return false;
} 
 flag = true;
bcd_emit(&bcd, "1");

crashpad::CrashpadClient::DumpWithoutCrash(context);

return true;
```

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -48,6 +91,167 @@ JNIEXPORT jint JNI_OnLoad(JavaVM *jvm, void *reserved) {
...
+        return false;
+    }
+
+    bool LocalUnwindingHandlerDumpWithoutCrash(int signum, siginfo_t *info, ucontext_t *context)
+    {
+        (void) signum;
+        (void) info;
+        (void) context;
+
+        thread_local bool flag;
```
konraddysput, line 188: This condition can be simplify ( check comment below)

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -48,6 +91,167 @@ JNIEXPORT jint JNI_OnLoad(JavaVM *jvm, void *reserved) {
...
+
+        return r;
+    }
+
+    static int
+    request_handler(pid_t tid)
+    {
+        // TODO revisit this logic
+        bool bun_initialized = bun_handle_init(&handle, BUN_BACKEND_LIBUNWINDSTACK);
+        if (!bun_initialized) {
```
konraddysput, line 131: we should be consistent with brackets. if statements above don't have brackets. Java layer has (and should have). Can we decide - do we want brackets or no?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -48,6 +91,167 @@ JNIEXPORT jint JNI_OnLoad(JavaVM *jvm, void *reserved) {
...
+            return -1;
+        }
+
+        bun_buffer buf = { buffer_child, BUFFER_SIZE };
+
+        auto written = bun_unwind_remote(&handle, &buf, tid);
+
+        (void) written;
+
+        return -1;
```
konraddysput, line 141: Are you sure request_handler should always (no matter what) return -1?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -321,4 +586,108 @@ Java_backtraceio_library_base_BacktraceBase_dumpWithoutCrash__Ljava_lang_String_
...
+
+    if (!bun_initialized) {
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Local client side unwinding initialization failed");
+        return false;
+    }
+
+    return true;
+}
+
+bool InitializeRemoteClientSideUnwinding(JNIEnv* env, jstring path) {
```
konraddysput, line 606: I still don't know why we can't pass a boolean variable to the initialize method and invoke this method from the initialize method. It will completely simplify API in my opinion. Why do we want to have separated functions for crash reporting api? We can end up in a situation where user won't know what and how to do stuff because our api will be to complex to understand or simply it will be too complex to understand via readme.

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -321,4 +586,108 @@ Java_backtraceio_library_base_BacktraceBase_dumpWithoutCrash__Ljava_lang_String_
...
+    struct bcd_config cf;
+    bcd_error_t e;
+
+    // Initialize a shared memory region.
+    static const char *buffer = static_cast<char *>(buffer_create());
+
+    // Initialize the BCD configuration file.
+    if (bcd_config_init(&cf, &e) == -1)
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
```
konraddysput, line 615: shouldn't we exit here?

konraddysput (commented):
Pushed few comments about api design.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
rqbacktrace: You're right, I think we can always just use the same path for the bcd socket file as we will use for the crashpad database directory. Any objection to this @KrzaQ ?

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: Will resolve, thank you!

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: That's a good point, what should we return on success for the bcd request_handler @KrzaQ ?

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: I think you're right, will make this change

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: That's a good point, will do!

example-app/src/main/cpp/native-lib.cpp
```diff
@@ -1,12 +1,21 @@
 #include <jni.h>
 #include <signal.h>
+#include <unistd.h>
 #include "backtrace-android.h"
 
+void * volatile always_null;
+
+void anotherCrash()
+{
+//    memset(always_null, 0x42, 1 << 20);
```
BartoszLitwiniuk, line 10: do we need that or can be removed?

BartoszLitwiniuk (approved):

example-app/src/main/cpp/native-lib.cpp
rqbacktrace: For native crashes it can be useful to test different ways of crashing. I will add a comment making this more clear, thank you!

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: It does look much better this way, thank you Konrad!

example-app/src/main/cpp/native-lib.cpp
konraddysput: @rqbacktrace  does it make sense to move this function to different .cpp file? Are you sure this should be a part of the `native-lib` source code? I think it belongs to "helpers" or "utils"?

konraddysput: Please keep in mind this code will end up in someone production code. We should be super careful with code that we're exposing/adding to libraries.

konraddysput:
Added @Dautery91 to pull request reviewers - based on the SDK ownership document, he and @jasoncdavis0 are responsible on the product side for backtrace-android releases.

backtrace-library/src/main/cpp/backtrace-native.cpp
rqbacktrace: @KrzaQ it's probably worth adding a comment to document this, I will do so now. But anyway the behavior is correct, thank you!

konraddysput (approved):

PR #28, opened 2021-08-05
Feature/increase client side unwinding robustness

rqbacktrace:
Support client side unwinding in NDK 17+ and increase robustness

### DIFF
--- a/backtrace-library/src/main/cpp/CMakeLists.txt
+++ b/backtrace-library/src/main/cpp/CMakeLists.txt
@@ -4,7 +4,7 @@
 # Sets the minimum version of CMake required to build the native library.
 
 cmake_minimum_required(VERSION 3.4.1)
-
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
 # Creates and names a library, sets it as either STATIC
 # or SHARED, and provides the relative paths to its source code.
 # You can define multiple libraries, and CMake builds them for you.
@@ -32,9 +32,13 @@ set_property(TARGET base PROPERTY IMPORTED_LOCATION ${PROJECT_SOURCE_DIR}/crashp
 # Crashpad Headers
 include_directories(${PROJECT_SOURCE_DIR}/crashpad-builds/headers/ ${PROJECT_SOURCE_DIR}/crashpad-builds/headers/third_party/mini_chromium/mini_chromium/)
 
-# Bun Libraries
-set(LIBUNWINDSTACK_ENABLED TRUE)
-add_subdirectory(libbun)
+if(NOT ANDROID_PLATFORM MATCHES "^android-([0-9]|1[0-9]|2[0-2])$")
+    # Bun Libraries
+    set(LIBUNWINDSTACK_ENABLED TRUE)
+    add_subdirectory(libbun)
+else()
+    message("API version < 23, client-side unwinding will be disabled")
+endif()
 
 # Searches for a specified prebuilt library and stores the path as a
 # variable. Because CMake includes system libraries in the search path by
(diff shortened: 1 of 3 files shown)


backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -133,8 +133,11 @@ namespace /* anonymous */
         buffer_fd = fd;
 
         r = buffer_reader(buffer_fd, PROT_READ|PROT_WRITE);
-
+#if DEBUG_BUFFER_CREATE
+        return nullptr;
```
pkhuong, line 137: why do we want to leak the `buffer_reader` when `-DDEBUG_BUFFER_CREATE`.

Also, IIUC, this `#define` enables failure injection in the `buffer_create` path? We should rename to reflect that.

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -133,8 +133,11 @@ namespace /* anonymous */
         buffer_fd = fd;
```
pkhuong, line 133: Nit: I don't know of any style that indents inside a `namespace`.

backtrace-library/src/main/cpp/CMakeLists.txt
```diff
@@ -32,9 +32,13 @@ set_property(TARGET base PROPERTY IMPORTED_LOCATION ${PROJECT_SOURCE_DIR}/crashp
 # Crashpad Headers
 include_directories(${PROJECT_SOURCE_DIR}/crashpad-builds/headers/ ${PROJECT_SOURCE_DIR}/crashpad-builds/headers/third_party/mini_chromium/mini_chromium/)
 
-# Bun Libraries
-set(LIBUNWINDSTACK_ENABLED TRUE)
-add_subdirectory(libbun)
+if(NOT ANDROID_PLATFORM MATCHES "^android-([0-9]|1[0-9]|2[0-2])$")
```
pkhuong, line 35: do you think it would be clearer to do `if(ANDROID_PLATFORM MATCHES "...") message() else () ...` ?

backtrace-library/src/main/cpp/CMakeLists.txt
```diff
@@ -55,14 +59,14 @@ find_library( # Sets the name of the path variable.
 # Specifies libraries CMake should link to your target library. You
 # can link multiple libraries, such as libraries you define in this
 # build script, prebuilt third-party libraries, or system libraries.
+list(APPEND LIBS backtrace-native)
+list(APPEND LIBS ${log-lib})
+list(APPEND LIBS crashpad_client)
+list(APPEND LIBS crashpad_util)
+list(APPEND LIBS base)
+
+if(NOT ANDROID_PLATFORM MATCHES "^android-([0-9]|1[0-9]|2[0-2])$")
```
pkhuong, line 68: can we use a named feature instead of repeating the condition here?

pkhuong: or unconditionally populate a list of libraries for `bun` (which could be an empty list) around L35?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -17,11 +17,13 @@
 #include "client/crash_report_database.h"
 #include "client/settings.h"
 
+#if __ANDROID_API__ >= 23
```
pkhuong, line 20: That's a lot of ifdef. Can we instead compile a file conditionally for API >= 23 (failing that, we could `#include` that file)?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -292,6 +279,7 @@ namespace /* anonymous */
         return m_pJniEnv;
     }
 
+#if __ANDROID_API__ >= 23
     bool InitializeLocalClientSideUnwinding(JNIEnv* env) {
```
pkhuong, line 286: Does the java side correctly gate calls to this function? Should be instead always define it, but have it fail for `API < 23`?

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -294,6 +295,10 @@ namespace /* anonymous */
...
         // Initialize a shared memory region.
         static const char *buffer = static_cast<char *>(buffer_create());
```
pkhuong, line 288: Why is this buffer `const char *` ? Did you mean `static char *const buffer` ?

pkhuong: do we have a specific naming style for globals? I think we should make it extra obvious that this is a `static` variable.

pkhuong: Actually, why do we want a `static` buffer?

pkhuong (changes requested):
I have a lot of questions, but mostly we should avoid sprinkling `#ifdef` everywhere

backtrace-library/src/main/cpp/backtrace-native.cpp
KrzaQ: It is an unfortunate side-effect of the IDE (Android Studio) enforcing its own style. Personally, I do not consider it worth the effort to ensure every contributor has a matching style settings for a minor part of the mainly-Java codebase.

rqbacktrace:
These changes are incorporated into NDK16b fixes

PR #30, opened 2021-08-31
Feature/ndk16bsupport

rqbacktrace:
For reviewers:
- This contains fixes for NDK 16b support using Breakpad
- This allows client side unwinding to work in all supported API/NDK levels (Crashpad backend only)
- 3,500 lines are just coming from the `cacert.h` file which you can probably ignore


vlussenburg (approved):
LGTM - mainly looked at human readable text not code

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -0,0 +1,69 @@
+#include "backend.h"
+#include "client-side-unwinding.h"
+#include <android/log.h>
+#include <atomic>
+#include <mutex>
+
+#if CRASHPAD_BACKEND
```
KrzaQ, line 7: We should use `#ifdef X` or `#if defined(X)`

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -0,0 +1,69 @@
...
+                jstring handler_path,
+                jobjectArray attributeKeys,
+                jobjectArray attributeValues,
+                jobjectArray attachmentPaths,
+                jboolean enableClientSideUnwinding,
+                jint unwindingMode) {
+    static std::once_flag initialize_flag;
+
+    std::call_once(initialize_flag, [&] {
+#if CRASHPAD_BACKEND
```
KrzaQ, line 27: Same, `#ifdef`

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -0,0 +1,69 @@
...
+#elif BREAKPAD_BACKEND
+    DumpWithoutCrashBreakpad(message, set_main_thread_as_faulting_thread);
+#else
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
+                        "DumpWithoutCrash not supported on this backend");
+#endif
+}
+
+void AddAttribute(jstring key, jstring value) {
+#if CRASHPAD_BACKEND
```
KrzaQ, line 60: Same as above

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+bool InitializeCrashpad(jstring url,
+                        jstring database_path,
+                        jstring handler_path,
+                        jobjectArray attributeKeys,
+                        jobjectArray attributeValues,
+                        jobjectArray attachmentPaths,
+                        jboolean enableClientSideUnwinding,
+                        jint unwindingMode) {
+    using namespace crashpad;
```
KrzaQ, line 18: We barely use the names from inside this namespace, maybe just use fully-qualified names instead?

backtrace-library/src/main/cpp/client-side-unwinding.cpp
```diff
@@ -0,0 +1,387 @@
...
+    if (initialized) {
+        __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
+                            "Client side unwinding needs to be enabled BEFORE crashpad initialization");
+        return false;
+    }
+
+    unwinding_mode = static_cast<UnwindingMode>((int) unwindingMode);
+
+    switch (unwinding_mode) {
+        case UnwindingMode::LOCAL:
```
KrzaQ, line 270: I'd add a `/* fallthrough */` comment or use the [`[[fallthrough]]`](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) attribute.

backtrace-library/src/main/cpp/include/breakpad-backend.h
```diff
@@ -0,0 +1,48 @@
...
+#include <unordered_map>
+#include <string>
+
+static constexpr size_t MAX_UPLOADS_PER_RUN = 3;
+
+extern std::string thread_id;
+extern std::atomic_bool initialized;
+extern std::mutex attribute_synchronization;
+
+static google_breakpad::ExceptionHandler* eh;
```
KrzaQ, line 20: We should not declare static variables in headers, ever.

backtrace-library/src/main/cpp/include/crashpad-backend.h
```diff
@@ -0,0 +1,31 @@
...
+#include "client/crashpad_info.h"
+#include "client/crash_report_database.h"
+#include "client/settings.h"
+
+#include "client-side-unwinding.h"
+
+#include <jni.h>
+
+// crashpad client
+using namespace base;
```
KrzaQ, line 15: We should strive to avoid using `using namespace`, and never do this in the global namespace in a header.

If we _must_ use `using namespace`, at the very most, do it at a function/class/inner namespace level.

backtrace-library/src/main/cpp/include/crashpad-backend.h
```diff
@@ -0,0 +1,31 @@
...
+static crashpad::CrashpadClient *client;
```
KrzaQ, line 16: Same comment about `static` as in `breakpad-backend.h`

example-app/src/main/cpp/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ project("myapplication")
 # or SHARED, and provides the relative paths to its source code.
 # You can define multiple libraries, and CMake builds them for you.
 # Gradle automatically packages shared libraries with your APK.
-set(CMAKE_CXX_FLAGS -fno-omit-frame-pointer)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer -std=c++11")
```
KrzaQ, line 16: We should use [`target_compile_options`](https://cmake.org/cmake/help/latest/command/target_compile_options.html) instead. Assuming that we target CMake versions that support it, but I think we do.

backtrace-library/src/main/cpp/CMakeLists.txt
```diff
@@ -4,37 +4,90 @@
...
-add_library(base STATIC IMPORTED)
-set_property(TARGET base PROPERTY IMPORTED_LOCATION ${PROJECT_SOURCE_DIR}/crashpad-builds/${ANDROID_ABI}/third_party/mini_chromium/mini_chromium/base/libbase.a)
-
-# Crashpad Headers
-include_directories(${PROJECT_SOURCE_DIR}/crashpad-builds/headers/ ${PROJECT_SOURCE_DIR}/crashpad-builds/headers/third_party/mini_chromium/mini_chromium/)
-
-# Bun Libraries
-set(LIBUNWINDSTACK_ENABLED TRUE)
-add_subdirectory(libbun)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
KrzaQ, line 7: We should use [`target_compile_options`](https://cmake.org/cmake/help/latest/command/target_compile_options.html) instead. Assuming that we target CMake versions that support it, but I think we do.

KrzaQ (changes requested):

backtrace-library/src/main/cpp/backends/backend.cpp
rqbacktrace: Will change across files, thank you!

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
rqbacktrace: Agreed, will do!

backtrace-library/src/main/cpp/client-side-unwinding.cpp
rqbacktrace: We can't assume we have C++17 so I will add the comment to make it clear!

backtrace-library/src/main/cpp/include/breakpad-backend.h
rqbacktrace: Ah sorry, will move these to implementations to avoid having multiple copies of the static variables

backtrace-library/src/main/cpp/include/crashpad-backend.h
rqbacktrace: Sounds good, will just fully qualify these symbols

KrzaQ (approved):

vlussenburg (approved):

PR #31, opened 2021-09-02
Upgrade Android Gradle plugin so we can build with NDK 22

rqbacktrace:
For reviewers:

This change will upgrade Android Gradle plugin so we can build with NDK 22, and so we can work with the latest versions of Android Studio

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Backtrace Android Release Notes
 
-## Version 3.5.0 - 02.09.2021
+## Version 3.4.0 - 07.09.2021
 - Added support for NDK 22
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Backtrace Android Release Notes
 
+## Version 3.5.0 - 02.09.2021
```
konraddysput, line 3: 3.4.0

BartoszLitwiniuk (approved):
Please fix library version in the CHANGELOD.md and apart from that I don't see other problems. Good job

rqbacktrace:
Thank you for approval! I set version as 3.5.0 because I thought we might release another version earlier but it never happened so this is now 3.4.0 release!

PR #35, opened 2021-10-20
Feature/unify attributes

rqbacktrace:
**What?**
This change unifies BacktraceAttributes with DeviceAttributesHelper. This change also will include DeviceAttributesHelper attributes in Crashpad crashes.
**Why?**
We don't have a case where we only want BacktraceAttributes without DeviceAttributesHelper, and previously in Crashpad crashes we were ignoring DeviceAttributesHelper attributes. Unifying these attribute getters will prevent losing attributes by accident in the future.
**How?**
BacktraceAttributes will always get DeviceAttributesHelper attributes in constructor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ hs_err_pid*
 #NDK
 obj/
 .externalNativeBuild
+**/.cxx


BartoszLitwiniuk (approved):
good job

PR #37, opened 2021-11-16
Provide support for API 16+

rqbacktrace:
These changes will allow us to support Android API levels 16+


vlussenburg (approved):

konraddysput (approved):
It looks good on my side! but I think @KrzaQ should make a final call here

KrzaQ (approved):

PR #39, opened 2021-11-22
Prevent ANR duplicates

konraddysput:
# Why

In the past when ANR took longer than expected, backtrace-android might report multiple ANRs when the library shouldn't. This diff prevents duplicates.

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/watchdog/BacktraceANRWatchdog.java
+++ b/backtrace-library/src/main/java/backtraceio/library/watchdog/BacktraceANRWatchdog.java
@@ -128,7 +128,7 @@ public class BacktraceANRWatchdog extends Thread {
                 continue;
             }
             if(reported) {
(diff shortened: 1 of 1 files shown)


backtrace-library/src/main/java/backtraceio/library/watchdog/BacktraceANRWatchdog.java
```diff
@@ -125,6 +127,11 @@ public void run() {
                         "is on and connected debugger");
                 continue;
             }
+            if(reported) {
+                // skiping, because we already reported an ANR report for current ANR
```
vlussenburg, line 131: skipPing

vlussenburg (changes requested):

rqbacktrace (approved):
Looks good to me! While we are in ANR state, we will only send one report. We won't send another report until we have newly entered ANR state again.

vlussenburg (approved):

PR #40, opened 2022-01-06
Enable GitHub actions workflows

rqbacktrace:
Enable GitHub actions for unit tests which should run every day and on pull request and push to master.


.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
+name: Run Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron:  '0 0 * * *'
```
BartoszLitwiniuk, line 9: is it necessary to run tests everyday if tests are not related with any other components/database/api etc? maybe it should be just run on commit and on PR?

rqbacktrace: I agree, what do you think @vlussenburg? In the ticket it asks for daily test runs but it might not be necessary

.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
+name: Run Tests
```
BartoszLitwiniuk, line 1: If we want to run tests by github actions I think that it is worth to remove/disable that in Travis CI https://travis-ci.org/github/backtrace-labs/backtrace-android

rqbacktrace: We can fully replace Travis CI. [GitHub Actions is free for private repos](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions). Do you agree @vlussenburg to commit to migrating fully to GitHub Actions and deprecating Travis?

BartoszLitwiniuk (approved):
Good job!

.github/workflows/test.yml
vlussenburg: The rationale for daily unit tests is the possibility of them magically stopping to work. IE maybe a dependency is out of date, maybe there's a weird dependency on an external service, maybe a unit test fails every 5th of the month. It's (practically) free to run tests daily and it make sure that when an engineer needs to make a change (remember we don't touch our SDKs weekly or even monthly sometimes) that they are sure it's still working instead of having to spend hours fixing unit tests that stopped working.

Please keep it as such :)

.github/workflows/test.yml
vlussenburg: Yes

vlussenburg (approved):

vlussenburg:
Nice!!

konraddysput (approved):
Looks great! Nothing more to add on my side.

.github/workflows/test.yml
rqbacktrace: Will keep it as such, thank you @vlussenburg !

.github/workflows/test.yml
rqbacktrace: Sounds good! I'll make the README badge change here in this pull request as well

PR #48, opened 2022-07-07
Update BacktraceQueueFileHelper.java

vlussenburg:
There's a situation where the breadcrumb file gets erased completely because `usedBytes` isn't updated in the while loop


BartoszLitwiniuk (approved):
lgtm

PR #53, opened 2022-07-29
Feature/cmake crashpad

vlussenburg:


KrzaQ:
The changes themselves look good to me. 👍

I would like to not spam the history, though, so at least squash the "update text.yml" commits together.

KrzaQ (approved):
As said before, squash is necessary.

example-app/src/androidTest/java/backtraceio/backtraceio/ExampleInstrumentedTest.java
```diff
@@ -16,11 +24,46 @@
...
+        onView(withId(R.id.handledException)).perform(click());
+    }
+
+    @Test
+    public void anr() {
+        onView(withId(R.id.anr)).perform(click());
+    }
+
+    // Will break build, obviously.
+    //@Test
```
BartoszLitwiniuk, line 65: do we need this comment `//@Test`?

example-app/src/androidTest/java/backtraceio/backtraceio/ExampleInstrumentedTest.java
```diff
@@ -16,11 +24,46 @@
...
+    public void nativeCrash() {
```
BartoszLitwiniuk, line 66: is this method called anywhere?

.github/workflows/test.yml
```diff
@@ -10,41 +10,118 @@ on:
...
+          morgue --token=${{ secrets.BACKTRACE_API_TOKEN }} --endpoint="https://yolo.sp.backtrace.io" list yolo/android-library --select timestamp --filter=timestamp,greater-than,`date +%s%3N -d "15 minutes ago"` || true
+
+#           if [ $(wc -l < crash.csv) -gt 0 ]; then
+#               echo "Success"
+#           else
+#               exit 1
+#           fi
+
+    #        runs-on: macos-latest # necessary for reactivecircus/android-emulator-runner@v2
+    #     - name: Emulator test
```
BartoszLitwiniuk, line 118: do we need it?

example-app/src/androidTest/java/backtraceio/backtraceio/ExampleInstrumentedTest.java
vlussenburg: Good catch. it’s left there to easily turn on the test. I need to figure out how to run the test (which crashes the app) without crashing the test. Or accept that the test fails but then we need to add a custom annotation that prevent it from running on “normal runs”.

example-app/src/androidTest/java/backtraceio/backtraceio/ExampleInstrumentedTest.java
vlussenburg: See the above.

.github/workflows/test.yml
vlussenburg: Maybe. Apilevel 16 is not supported on Sauce, which would be nice to test on. So we may have to boot an old school emulator.

PR #55, opened 2022-08-18
Crashpad attachment syntax changed

vlussenburg:


KrzaQ (approved):
This is exactly what we discussed.

PR #61, opened 2022-08-20
Upgrade gson version

vlussenburg:
https://ossindex.sonatype.org/vulnerability/sonatype-2021-1694?component-type=maven&component-name=com.google.code.gson%2Fgson&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0


BartoszLitwiniuk (approved):
lgtm

PR #63, opened 2022-09-06
Androidx migration

BartoszLitwiniuk:


vlussenburg (approved):
LGTM

backtrace-library/build.gradle
```diff
@@ -11,16 +11,16 @@ androidGitVersion {
 
 android {
 
-    compileSdkVersion 30
+    compileSdkVersion 32
```
vlussenburg, line 14: Could you leave a comment why minAPI is 32 so we’ll remember in the future?

BartoszLitwiniuk: minSdkVersion is `21`

PR #70, opened 2022-10-06
Upgrade to Android 13 / SDK 33

vlussenburg:


vlussenburg:
@BartoszLitwiniuk I saw the message:

```
WARNING:Software Components will not be created automatically for Maven publishing from Android Gradle Plugin 8.0. To opt-in to the future behavior, set the Gradle property android.disableAutomaticComponentCreation=true in the `gradle.properties` file or use the new publishing DSL.
WARNING:We recommend using a newer Android Gradle plugin to use compileSdk = 33
```

WDYT?

BartoszLitwiniuk (approved):
I've check a stackoverflow, docs and pages, few links below:
https://developer.android.com/about/versions/13/behavior-changes-all
https://developer.android.com/about/versions/13/reference/compat-framework-changes
https://developer.android.com/about/versions/13/features

I think there are no contraindications to do this

BartoszLitwiniuk:
> @BartoszLitwiniuk I saw the message:
> 
> ```
> WARNING:Software Components will not be created automatically for Maven publishing from Android Gradle Plugin 8.0. To opt-in to the future behavior, set the Gradle property android.disableAutomaticComponentCreation=true in the `gradle.properties` file or use the new publishing DSL.
> WARNING:We recommend using a newer Android Gradle plugin to use compileSdk = 33
> ```
> 
> WDYT?

Maybe it's worth to upgrade gradle before merging this?

PR #71, opened 2022-10-12
Feature/crash loop detection

vlussenburg:


konst-sauce:
@konraddysput @KrzaQ @vlussenburg - please review my latest changes and comment, if OK - feel free to approve :)

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -82,4 +82,29 @@ void Disable() {
                         "Disable not supported on this backend");
 #endif
 }
+
+jboolean EnableCrashLoopDetectionBackend() {
+#ifdef CRASHPAD_BACKEND
+    return EnableCrashLoopDetectionCrashpad();
+#elif BREAKPAD_BACKEND
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "EnableCrashLoopDetection not supported on this backend");
```
konraddysput, line 90: I think we need to make things more obvious for unity and other tools that will be using it. I think we need to return false here to clearly show it doesn't work

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -19,6 +23,9 @@ bool InitializeCrashpad(jstring url,
                         jobjectArray attachmentPaths,
                         jboolean enableClientSideUnwinding,
                         jint unwindingMode) {
+
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Initialize Crashpad Start");
```
konraddysput, line 27: I think we dont need this error log

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -120,9 +127,19 @@ bool InitializeCrashpad(jstring url,
...
-    initialized = client->StartHandlerAtCrash(handler, db, db, backtraceUrl, attributes,
-                                              arguments);
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Crashpad created.");
+    client->EnableCrashLoopDetection();
+
+    // Get consecutive crashes count BEFORE any handler started,
+    // as it writes extra line into CSV, what leads to getting 0 for each next ConsecutiveCrashesCount call
+    consecutive_crashes_count = crashpad::CrashpadClient::ConsecutiveCrashesCount(db);
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Crashpad initialize - ConsecutiveCrashesCount %d", consecutive_crashes_count);
+//    initialized = client->StartHandler(handler, db, db, backtraceUrl, attributes, arguments, false, false);
```
konraddysput, line 137: commented code - we should remove it

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -219,10 +236,28 @@ void ReEnableCrashpad() {
...
         database->GetSettings()->SetUploadsEnabled(true);
         disabled = false;
     }
 }
+
+bool EnableCrashLoopDetectionCrashpad() {
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Inside EnableCrashLoopDetectionCrashpad");
+    if (client != nullptr) {
+        return client->EnableCrashLoopDetection();
+    } else {
```
konraddysput, line 252: does it make sense to print warning the client is not available? We need to be clear in client logs to show what doesn't work. We need to show it fast and hard

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -135,4 +135,22 @@ Java_backtraceio_library_BacktraceDatabase_disable(JNIEnv *env, jobject thiz) {
     Disable();
 }
 
+extern "C"
+JNIEXPORT jboolean JNICALL
+Java_backtraceio_library_base_BacktraceBase_EnableCrashLoopDetectionBacktrace(JNIEnv *env, jclass clazz) {
```
konraddysput, line 140: This shouldn't belong to the BacktraceBase class - instead, it should belong to BacktraceDatabase. Why? To enable and do any operation via Crashpad we need a database. If we do not have the database, this code should fail. There is no reason to run it in the client if the database is unavailable. It belongs to the database domain

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -49,10 +49,30 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.activity_main);
 
         // Set this value in your local.properties
-        if (BuildConfig.BACKTRACE_SUBMISSION_URL != null) {
-            backtraceClient = initializeBacktrace(BuildConfig.BACKTRACE_SUBMISSION_URL);
-        }
+        String BACKTRACE_SUBMISSION_URL = "https://submit.backtrace.io/konst-ryab/04aecb6b3da05e8d83f2a27f9b5f41352ac792428c5e5f6a94ff58d43ee14e46/json";
```
konraddysput, line 52: this shouldn't be here

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -49,10 +49,30 @@ protected void onCreate(Bundle savedInstanceState) {
...
+//        if (BACKTRACE_SUBMISSION_URL != null) {
+            backtraceClient = initializeBacktrace(BACKTRACE_SUBMISSION_URL);
 
+        Context context = getApplicationContext();
```
konraddysput, line 56: this shouldn't be here

konraddysput (changes requested):
I think we need to move our interface to database instead of to client. We have also a lot of comments in our code that shouldn't be here.

vlussenburg:
Redoing this PR because binaries were pushed to the branch and had to filter them out

backtrace-library/src/main/cpp/backtrace-native.cpp
konst-sauce: Moved to BacktraceDatabase

PR #74, opened 2022-10-17
Handle corrupt or empty database records

selalipop:
In cases where a crash record file is corrupted or empty, `BacktraceDatabaseRecord.readFromFile` may return null, causing a crash when the result accessed

This PR adds handling for this case by treating null return values as an invalid record

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
@@ -415,7 +415,7 @@ public class BacktraceDatabase implements Database {
 
         for (File file : files) {
             BacktraceDatabaseRecord record = BacktraceDatabaseRecord.readFromFile(file);
-            if (!record.valid()) {
+            if (record == null || !record.valid()) {
                 record.delete();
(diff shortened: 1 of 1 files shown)


konraddysput (approved):

selalipop:
Yes it will! Thank you for the catch 😁 

And I'll come up with some test cases for this

BartoszLitwiniuk (changes requested):
Please check my comments

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -415,7 +415,7 @@ private void loadReports() {
 
         for (File file : files) {
             BacktraceDatabaseRecord record = BacktraceDatabaseRecord.readFromFile(file);
-            if (!record.valid()) {
+            if (record == null || !record.valid()) {
```
BartoszLitwiniuk, line 418: Your solution will cause NullPointerException if record will be null

BartoszLitwiniuk:
I'm closing this PR and I will open new one with proper fix and unit test

PR #80, opened 2022-11-29
Feature/crash loop detection

vlussenburg:
Reopening https://github.com/backtrace-labs/backtrace-android/pull/71

### DIFF
--- a/backtrace-library/src/main/cpp/backends/backend.cpp
+++ b/backtrace-library/src/main/cpp/backends/backend.cpp
@@ -82,4 +82,29 @@ void Disable() {
                         "Disable not supported on this backend");
 #endif
 }
+
+jboolean EnableCrashLoopDetectionBackend() {
+#ifdef CRASHPAD_BACKEND
+    return EnableCrashLoopDetectionCrashpad();
+#elif BREAKPAD_BACKEND
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "EnableCrashLoopDetection not supported on this backend");
+#endif
+}
+
+jboolean IsSafeModeRequiredBackend() {
+#ifdef CRASHPAD_BACKEND
+    return IsSafeModeRequiredCrashpad();
+#elif BREAKPAD_BACKEND
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "IsSafeModeRequired not supported on this backend");
+#endif
+}
+
+jint ConsecutiveCrashesCountBackend() {
+#ifdef CRASHPAD_BACKEND
+    return ConsecutiveCrashesCountCrashpad();
+#elif BREAKPAD_BACKEND
+    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "ConsecutiveCrashesCount not supported on this backend");
+#endif
+}
+
 }
--- a/backtrace-library/src/main/cpp/backends/backend.cpp
+++ b/backtrace-library/src/main/cpp/backends/backend.cpp
@@ -83,27 +83,27 @@ void Disable() {
 #endif
 }
 
-jboolean EnableCrashLoopDetectionBackend() {
+bool EnableCrashLoopDetectionBackend() {
 #ifdef CRASHPAD_BACKEND
     return EnableCrashLoopDetectionCrashpad();
 #elif BREAKPAD_BACKEND
-    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "EnableCrashLoopDetection not supported on this backend");
+    return false;
 #endif
 }
 
-jboolean IsSafeModeRequiredBackend() {
+bool IsSafeModeRequiredBackend() {
 #ifdef CRASHPAD_BACKEND
     return IsSafeModeRequiredCrashpad();
 #elif BREAKPAD_BACKEND
-    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "IsSafeModeRequired not supported on this backend");
+    return false;
 #endif
 }
 
-jint ConsecutiveCrashesCountBackend() {
+int ConsecutiveCrashesCountBackend() {
 #ifdef CRASHPAD_BACKEND
     return ConsecutiveCrashesCountCrashpad();
 #elif BREAKPAD_BACKEND
-    __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "ConsecutiveCrashesCount not supported on this backend");
+    return 0;
 #endif
 }
 
--- a/backtrace-library/src/main/cpp/backends/backend.cpp
+++ b/backtrace-library/src/main/cpp/backends/backend.cpp
@@ -83,7 +83,7 @@ void Disable() {
 #endif
 }
 
-bool EnableCrashLoopDetectionBackend() {
+bool EnableCrashLoopDetection() {
 #ifdef CRASHPAD_BACKEND
     return EnableCrashLoopDetectionCrashpad();
 #elif BREAKPAD_BACKEND
@@ -91,7 +91,7 @@ bool EnableCrashLoopDetectionBackend() {
 #endif
 }
 
-bool IsSafeModeRequiredBackend() {
+bool IsSafeModeRequired() {
 #ifdef CRASHPAD_BACKEND
     return IsSafeModeRequiredCrashpad();
 #elif BREAKPAD_BACKEND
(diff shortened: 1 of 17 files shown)


konraddysput (changes requested):

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -3,6 +3,8 @@
 #include <jni.h>
 #include <libgen.h>
 
+//#include "client/crashpad_client.h"
```
vlussenburg, line 6: Remove

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -49,10 +49,30 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.activity_main);
 
         // Set this value in your local.properties
-        if (BuildConfig.BACKTRACE_SUBMISSION_URL != null) {
-            backtraceClient = initializeBacktrace(BuildConfig.BACKTRACE_SUBMISSION_URL);
-        }
+        String BACKTRACE_SUBMISSION_URL = "https://submit.backtrace.io/konst-ryab/04aecb6b3da05e8d83f2a27f9b5f41352ac792428c5e5f6a94ff58d43ee14e46/json";
+//        if (BACKTRACE_SUBMISSION_URL != null) {
```
vlussenburg, line 53: Restore to original

backtrace-library/src/main/cpp/backtrace-native.cpp
```diff
@@ -135,4 +135,22 @@ Java_backtraceio_library_BacktraceDatabase_disable(JNIEnv *env, jobject thiz) {
     Disable();
 }
 
+extern "C"
+JNIEXPORT jboolean JNICALL
+Java_backtraceio_library_base_BacktraceBase_EnableCrashLoopDetectionBacktrace(JNIEnv *env, jclass clazz) {
```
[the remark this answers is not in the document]
konst-sauce: OK

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
konst-sauce: OK

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -82,4 +82,29 @@ void Disable() {
                         "Disable not supported on this backend");
 #endif
 }
+
+bool EnableCrashLoopDetectionBackend() {
```
konraddysput, line 86: it doesn't make sense to add `Backend` suffix here. It doesn't provide any useful information to the user but adds more confusion. What does the backend mean? I think it would be better if we rename method to `EnableCrashLoopDetection` and we follow the same pattern with the rest of those methods

KrzaQ (changes requested):
I think the change looks well overall, but some things should be more fleshed out before we release it to the public.

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -120,9 +123,12 @@ bool InitializeCrashpad(jstring url,
 
     // Start crash handler
     client = new crashpad::CrashpadClient();
+    client->EnableCrashLoopDetection();
```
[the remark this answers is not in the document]
konraddysput: i think we should - imo we should pass it and don't execute this path if we don't need it. By default we want have features opt-in

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -49,10 +52,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            backtraceClient = initializeBacktrace(BACKTRACE_SUBMISSION_URL);
         }
 
+        // Crash Loop Detector example
+        boolean isCLSafeModeReq = BacktraceDatabase.IsSafeModeRequired();
+        int crashesCount = BacktraceDatabase.ConsecutiveCrashesCount();
+        Log.i("BacktraceAndroid", "ConsecutiveCrashesCount: " + crashesCount);
+
+        View viewBackground = findViewById(R.id.viewBackground);
+        if(viewBackground != null) {
```
BartoszLitwiniuk, line 65: fix formatting

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -49,10 +52,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
         // Set this value in your local.properties
-        if (BuildConfig.BACKTRACE_SUBMISSION_URL != null) {
-            backtraceClient = initializeBacktrace(BuildConfig.BACKTRACE_SUBMISSION_URL);
+        if (BACKTRACE_SUBMISSION_URL != null) {
+            backtraceClient = initializeBacktrace(BACKTRACE_SUBMISSION_URL);
         }
 
+        // Crash Loop Detector example
+        boolean isCLSafeModeReq = BacktraceDatabase.IsSafeModeRequired();
```
konraddysput, line 60: this code breaks if crashpad is not initialized at all. We need to fix it before the final release. Not sure if this is the problem with the flow or in the crashpad internals

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -19,6 +21,7 @@ bool InitializeCrashpad(jstring url,
                         jobjectArray attachmentPaths,
                         jboolean enableClientSideUnwinding,
                         jint unwindingMode) {
+
```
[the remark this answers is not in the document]
konst-sauce: ok

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -120,9 +123,12 @@ bool InitializeCrashpad(jstring url,
 
     // Start crash handler
     client = new crashpad::CrashpadClient();
+    client->EnableCrashLoopDetection();
 
-    initialized = client->StartHandlerAtCrash(handler, db, db, backtraceUrl, attributes,
-                                              arguments);
+    // Get consecutive crashes count BEFORE any handler started,
+    // as it writes extra line into CSV, what leads to getting 0 for each next ConsecutiveCrashesCount call
+    consecutive_crashes_count = crashpad::CrashpadClient::ConsecutiveCrashesCount(db);
```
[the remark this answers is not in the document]
konst-sauce: We are using this value later when we call IsSafeModeRequired f.e. from MainActivity. 
From my tests followed that this value should be set before we set up any handler to crashpad, because then I always received 0.

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -219,10 +225,27 @@ void ReEnableCrashpad() {
...
+            __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android",
+                                "Crashpad database is null, this should not happen");
             return;
         }
         database->GetSettings()->SetUploadsEnabled(true);
         disabled = false;
     }
 }
+
+bool EnableCrashLoopDetectionCrashpad() {
```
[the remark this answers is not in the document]
konst-sauce: Using now, I removed enabling by client->EnableCrashLoopDetection(); after crashpad client initialisation from above's comment

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
konst-sauce: Okay - so this one, following our discussions, is what comes from crashpad internals.

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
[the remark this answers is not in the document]
konst-sauce: removed it - it is called from Example's MainActivity now.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
konraddysput: I don't think the formatting has been fixed. Can we run formatted?

konraddysput: `if(` for example - but it would be better if we format the whole file - in case of some other formatting issues that we don't see

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
KrzaQ: We cannot accept this PR without update on the crashpad side. I'm on it.

KrzaQ (changes requested):
We need to update the crashpad dependency for a fix that has yet to be done. I'm on it, but this must wait.

konraddysput (approved):

KrzaQ:
Please rebase the branch on the latest master branch.

PR #86, opened 2022-12-11
Add ANR breadcrumb and fix null db

BartoszLitwiniuk:


vlussenburg (approved):

vlussenburg:
Lgtm

backtrace-library/src/main/java/backtraceio/library/watchdog/BacktraceWatchdogShared.java
```diff
@@ -28,6 +28,7 @@ static void sendReportCauseBlockedThread(BacktraceClient backtraceClient, Thread
         if (onApplicationNotRespondingEvent != null) {
             onApplicationNotRespondingEvent.onEvent(exception);
         } else if (backtraceClient != null) {
+            backtraceClient.addBreadcrumb("ANR detected - thread is blocked");
```
BartoszLitwiniuk, line 31: Add unit test

BartoszLitwiniuk:
> Hmm Instrumentation run failed due to 'java.lang.SecurityException

I received information from @rick-bt that it's a bug in SauceLabs and we don't need to care about it if tests passed

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -330,7 +330,10 @@ public void setOnRequestHandler(RequestHandler requestHandler) {
      * @return true if we successfully enabled breadcrumbs
      */
     public boolean enableBreadcrumbs(Context context) {
-        return database.getBreadcrumbs().enableBreadcrumbs(context);
+        if (database != null && database.getBreadcrumbs() != null) {
```
konraddysput, line 333: can we move it to function?

konraddysput:
> > Hmm Instrumentation run failed due to 'java.lang.SecurityException
> 
> I received information from @rick-bt that it's a bug in SauceLabs and we don't need to care about it if tests passed

if you see anything similar please re-run the test. I already did that. We're working on solution now. IT should be fixed soon

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: good point, done

backtrace-library/src/main/java/backtraceio/library/watchdog/BacktraceWatchdogShared.java
BartoszLitwiniuk: done

BartoszLitwiniuk:
@vlussenburg @konraddysput could you review my PR again?

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -330,7 +330,10 @@ public void setOnRequestHandler(RequestHandler requestHandler) {
      * @return true if we successfully enabled breadcrumbs
      */
     public boolean enableBreadcrumbs(Context context) {
-        return database.getBreadcrumbs().enableBreadcrumbs(context);
+        if (isBreadcrumbsEnabled()) {
```
konraddysput, line 333: do i understand it corretly? if breadcrumbs are enabled, enable breadcrumbs otherwise return false? It sounds weird. Something is odd

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -546,4 +585,8 @@ private void addReportAttachments(BacktraceReport report) {
             }
         }
     }
+
+    private boolean isBreadcrumbsEnabled() {
+        return database != null && database.getBreadcrumbs() != null;
```
konraddysput, line 590: i would rather call it isBreadcrumbAvailable instead.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -460,7 +493,10 @@ public boolean addBreadcrumb(String message, BacktraceBreadcrumbType type, Backt
      * @return true if the breadcrumb was successfully added
      */
     public boolean addBreadcrumb(String message, Map<String, Object> attributes, BacktraceBreadcrumbType type) {
-        return database.getBreadcrumbs().addBreadcrumb(message, attributes, type);
+        if (isBreadcrumbsEnabled()) {
```
konraddysput, line 496: small nit:
it would be better if we return without nesting more code in the if statement. you can save one tab and overall code would be easier to read from the top to the bottom if you inverse the if statement
```
if (!isBreadcrumbsEnabled()) {
return false;
} 
return database.getBreadcrumbs().addBreadcrumb(message, attributes, type);
```

konraddysput (changes requested):
Left some comments - especially about the method that checks if the breadcrumbs are available - enabled in this context means something else. Can we please rename the method?

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: good point

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: now after refactor this method is called isBreadcrumbsAvailable so breadcrumbs can be enabled only when it's available

PR #89, opened 2023-02-13
Backtrace client api improvements

konraddysput:
This diff moves some API from BacktraceDatabase class to BacktraceClient. Most of the clients want to use BacktraceClient instead of Database. This diff adds more flexibility to the client to allow to use one of them

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
@@ -25,6 +25,28 @@ public class BacktraceClient extends BacktraceBase {
     private BacktraceANRWatchdog anrWatchdog;
 
     /**
+     * Enables Crash loop detection to verify if the application is in the crash loop.
+     */
+    public static boolean EnableCrashLoopDetection() {
+        return BacktraceDatabase.EnableCrashLoopDetection();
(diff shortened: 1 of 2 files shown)


backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
```diff
@@ -24,6 +24,28 @@ public class BacktraceClient extends BacktraceBase {
...
+    }
+    
+    /**
+     * Returns the number of consecutive application crashes 
+     */
+    public static int ConsecutiveCrashesCount() {
+        return BacktraceDatabase.ConsecutiveCrashesCount();
+    }
+
+
```
konst-sauce, line 48: I'm not sure I like one more layer in the overall call stack to CLD methods in the crashpad, it is 5th seems :)
But having it under Client is still good.

konst-sauce (approved):

backtrace-library/src/main/java/backtraceio/library/BacktraceClient.java
konraddysput: Please take a look into the pull request description - most customers will use client instead of the database object. It's easier and it's less to learn for them

PR #91, opened 2023-03-29
Update dependencies and sdk to v33

BartoszLitwiniuk:


konraddysput:
Hey @BartoszLitwiniuk 

@melekr should block the review. Also - can you confirm: can you see a crash on android app with android SDK 33?

KrzaQ:
Looking at the code changes, it looks minimal enough. But I'd like to have confirmation that we have tested both managed and native crashes.

melekr:
Minimal code changes & dependencies updates that should not affect stability.
lgtm pending crash tests!

BartoszLitwiniuk:
I've tested all types of exceptions and it looks good. This PR contains update of compileSdkVersion and targetSdkVersion in the example-app. Both versions are already updated in the library.

PR #94, opened 2023-05-08
Feature/integration verification tests

BartoszLitwiniuk:
This pull request adds a new module to backtrace-android that is responsible for building a query and sending it to API to retrieve errors that match the specified filters. The new module is named `coroner-client`, and it includes methods for building and sending a query to Coroner API.

In addition to building and sending a query to our API, the `coroner-client` also allows for advanced filtering, grouping, and attribute selection. The module builds a JSON query with specified filters, including filtering by timestamp, error type, and rxId. The module also allows for grouping of errors based on specific attributes and selecting additional attributes to be returned in the query results.

These added features will give greater control over the errors they retrieve from our API. This new module in the future will be used for creating next integration tests.

This pull request also includes an integration test that uses the `coroner-client` to ensure that the end-to-end integration is working correctly and errors are sent and processed. The integration test includes multiple test cases that cover various filter configurations.


vlussenburg (approved):

rick-bt:
Vincent lives! All is still right with the world!! :)

vlussenburg:
> Vincent lives! All is still right with the world!! :)

My Backtrace heart still beats strong ;)

coroner-client/src/main/java/backtraceio/coroner/CoronerHttpClient.java
```diff
@@ -0,0 +1,94 @@
...
+                resultJson,
+                CoronerApiResponse.class);
+    }
+
+    private static String getResponseMessage(final HttpURLConnection urlConnection) throws IOException {
+        LOGGER.log(Level.INFO, "Reading response from HTTP request");
+
+        final InputStream inputStream = getInputStream(urlConnection);
+
+        final BufferedReader br = new BufferedReader(new InputStreamReader(
```
konraddysput, line 52: can we use full names?

coroner-client/src/main/java/backtraceio/coroner/query/CoronerQueryBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+    private final int LIMIT = 1;
+
+    public String buildRxIdGroup(final String filters, final List<String> headFolds) {
+        return this.build(CoronerQueryFields.RXID, filters, headFolds);
+    }
+
+    private String build(final String groupName, final String filters, final List<String> headFolds) {
+        final String folds = joinHeadFolds(headFolds);
+
+        return "{" +
```
konraddysput, line 18: why you building a json manually and not via library. What would happen if you get a character that breaks the JSON ?

coroner-client/src/main/java/backtraceio/coroner/response/CoronerResponseGroup.java
```diff
@@ -0,0 +1,25 @@
+package backtraceio.coroner.response;
+
+import java.util.List;
+
+public class CoronerResponseGroup {
+    private static final Integer EXPECTED_NUMBER_OF_EL = 3;
```
konraddysput, line 6: what is EL - can we use full name?

coroner-client/src/main/java/backtraceio/coroner/response/CoronerResponseGroup.java
```diff
@@ -0,0 +1,25 @@
...
+    @SuppressWarnings("unused")
+    private final String groupIdentifier;
+    private final List<Object> values;
+
+    public CoronerResponseGroup(final List<Object> obj) throws IllegalArgumentException {
+        if (obj == null || obj.size() != EXPECTED_NUMBER_OF_EL) {
```
konraddysput, line 12: what if some attributes already exists, or we have less attributes than 3 (which i believe is fine)

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -171,14 +176,24 @@ public void anr(View view) throws InterruptedException {
         Thread.sleep(anrTimeout + 2000);
     }
 
-    public void enableBreadcrumbs(View view) {
+    public void enableBreadcrumbs(View view) throws Exception {
+        Context appContext = view.getContext().getApplicationContext();
+        if (backtraceClient == null) {
+            throw new Exception("BacktraceClient is null");
```
konraddysput, line 182: I think it would be better to write a message like "Backtrace integration is not initialized"

coroner-client/src/main/java/backtraceio/coroner/response/CoronerResponseGroup.java
BartoszLitwiniuk: done

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
BartoszLitwiniuk: done

coroner-client/src/main/java/backtraceio/coroner/response/CoronerResponseGroup.java
BartoszLitwiniuk: Here there is a check if in the response below, element of values arrays contains 3 elements inside, in this case "*", array of attributes and 818. It's not a check if in the attributes array we have less than 3 elements. It's part of checking correctness of the response format.

```
"values": [
      [
        "*",
        [
          [
            "DumpWithoutCrash"
          ],
          [
            "{\"frame\":[\"libbacktrace-native.so\",\"libbacktrace-native.so\",\"libart.so\",\"libc.so\",\"libbacktrace-native.so\",\"libbacktrace-native.so\",\"libart.so\",\"libbacktrace-native.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libc.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\",\"libart.so\"]}"
          ],
          [
            "71baf2e3-0fe3-48fa-ba18-d2b3d98d1e75"
          ],
          [
            "exception"
          ]
        ],
        818
      ]
    ],
```

coroner-client/src/main/java/backtraceio/coroner/CoronerClient.java
```diff
@@ -0,0 +1,69 @@
...
+    public CoronerClient(HttpClient httpClient) {
+        this.coronerHttpClient = httpClient;
+        this.coronerQueries = new CoronerQueries();
+    }
+
+    public CoronerResponse rxIdFilter(final String rxId) throws CoronerResponseException, CoronerHttpException, IOException  {
+        return this.rxIdFilter(rxId, new ArrayList<>());
+    }
+
+    public CoronerResponse rxIdFilter(final String rxId, final List<String> customAttributes) throws CoronerResponseException, CoronerHttpException, IOException {
```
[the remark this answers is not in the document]
BartoszLitwiniuk: CoronerClient is a facade which provides concrete functionality needed. Builder pattern is used in classes `CoronerFiltersBuilder` and partially in `CoronerQueryBuilder`. The current implementation provides all currently required functionality, we will extend the builder options of these classes in case of future needs.

coroner-client/src/main/java/backtraceio/coroner/query/CoronerQueryBuilder.java
BartoszLitwiniuk: Let's keep it as it is because this query syntax is quite problematic like e.g. arrays of arrays with index dependent elements (check snippet below). We will use it in recent future only in integration tests so I don't want to spend more time on that. Of course you're right that it will be good to have it in the future.

```
    "guid": [
                [
                    "equal",
                    "2e55f002-32c6-56e4-9210-b54ac25be583"
                ]
            ],
```

konraddysput: Next time, can you please don't resolve comments that don't address the resolution to the question? On Github it's super hard to find out what did we change and what we didn't - later I'm not sure if we finished something or not. 

The example above won't be serialized to `string[][]`? We know where can we find information. 


if you want to stay with the custom JSON, how you're dealing with escapes or non cultural characters - use emoji for example. 

This is a matter of time when it will break because of something super trivial

coroner-client/src/main/java/backtraceio/coroner/CoronerClient.java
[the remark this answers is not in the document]
konraddysput: Let's make it clear - you can have one method instead of concrete functions and you will have the same - or actually better, you will have less code. 

In my opinion - we can easily change that, but I don't treat it as a blocker for this pull request.

coroner-client/src/main/java/backtraceio/coroner/query/CoronerQueryBuilder.java
BartoszLitwiniuk: Hi Konrad, I've decided to address your issue and now query JSON is build using library instead of string concatenation.

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
```diff
@@ -112,48 +114,51 @@ private BacktraceClient initializeBacktrace(final String submissionUrl) {
...
+                int magicWandIndex = findEquipmentIndex(myWarriorArmor, "Magic Wand");
+                // I don't need a Magic Wand, I am a warrior
+                removeEquipment(myWarriorArmor, magicWandIndex);
+                // Where was that magic wand again?
+                equipItem(myWarriorArmor, magicWandIndex);
+            } catch (IndexOutOfBoundsException e) {
+                throw new IndexOutOfBoundsException("Invalid index of selected element!");
+            }
+        } catch (IndexOutOfBoundsException e) {
+            backtraceClient.send(new BacktraceReport(e), this.listener);
```
konraddysput, line 155: wait, why? I see we want to show inner errors case here, however, can we make it more prominent?

coroner-client/src/test/java/backtraceio/coroner/serialization/GsonWrapperTest.java
```diff
@@ -0,0 +1,45 @@
...
...iew.performClickInternal\",\"android.view.View.access$3600\",\"android.view.View$PerformClick.run\",\"androidx.test.espresso.base.Interrogator.loopAndInterrogate\",\"androidx.test.espresso.base.UiControllerImpl.loopUntil\",\"androidx.test.espresso.base.UiControllerImpl.injectMotionEvent\",\"androidx.test.espresso.action.MotionEvents.sendUp\",\"androidx.test.espresso.action.Tap.sendSingleTap\",\"androidx.test.espresso.action.Tap.-$$Nest$smsendSingleTap\",\"androidx.test.espresso.action.Tap$1.sendTap\",\"androidx.test.espresso.action.GeneralClickAction.perform\",\"androidx.test.espresso.ViewInteraction$SingleExecutionViewAction.perform\",\"androidx.test.espresso.ViewInteraction.doPerform\",\"androidx.test.espresso.ViewInteraction.-$$Nest$mdoPerform\",\"androidx.test.espresso.ViewInteraction$1.call\",\"java.util.concurrent.FutureTask.run\",\"android.os.Handler.handleCallback\",\"android.os.Handler.dispatchMessage\",\"android.os.Looper.loop\",\"android.app.ActivityThread.main\",\"java.lang.reflect.Method.invoke\",\"com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run\",\"com.android.internal.os.ZygoteInit.main\"]}", getResponseGroupAttributeValue(responseGroup.getAttribute(1)));
```
konraddysput, line 41: can we move it to a file?

konraddysput (approved):

coroner-client/src/test/java/backtraceio/coroner/serialization/GsonWrapperTest.java
BartoszLitwiniuk: done

example-app/src/main/java/backtraceio/backtraceio/MainActivity.java
BartoszLitwiniuk: it's bugfix for integration test purpose.

in shortcut: different SDKs/devices contains different default message in IndexOutOfBoundsException so if we want to verify in the integration test if certain exception was thrown and we would like to verify error message we need to have it always in the same form and with the same value 

image

If you want I can keep it as it is or try to rearrange the method to e.g. throw another type of exception which will not cause problems like this

coroner-client/src/main/java/backtraceio/coroner/query/CoronerValueFilter.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Override
+    public String toString() {
+        return this.get().toString();
+    }
+
+    public JsonElement get() {
+        JsonArray result = new JsonArray();
+
+        result.add(operator.toString());
```
konraddysput, line 23: wouldn't it be better if you pass string to constructor instead of an object? Why class implementation should know about converting operator to string?

konraddysput (approved):

coroner-client/src/main/java/backtraceio/coroner/CoronerHttpClient.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private static String getResponseMessage(final HttpURLConnection urlConnection) throws IOException {
+        LOGGER.log(Level.INFO, "Reading response from HTTP request");
+
+        final InputStream inputStream = getInputStream(urlConnection);
+
+        final BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(
+                inputStream));
+
+        final StringBuilder responseSB = new StringBuilder();
```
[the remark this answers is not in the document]
BartoszLitwiniuk: done

coroner-client/src/main/java/backtraceio/coroner/query/CoronerValueFilter.java
konraddysput: java thing...

BartoszLitwiniuk: unfortunately it's a trick to have a string enum in Java

PR #95, opened 2023-05-23
Handle invalid server URL

konraddysput:
# Why

This diff handles invalid submission URL/token provided by the user to the BacktraceCredentials. In some situations when we need to :
* create a fake client for testing purposes,
* use database connection to determine when to send reports

the URL provided by the user might not be valid. This diff handles a situation when the user doesn't pass a valid credentials to the BacktraceClient

### DIFF
--- a/backtrace-library/src/androidTest/java/backtraceio/library/BacktraceCredentialsTest.java
+++ b/backtrace-library/src/androidTest/java/backtraceio/library/BacktraceCredentialsTest.java
@@ -1,8 +1,10 @@
 package backtraceio.library;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import android.net.Uri;
+
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


backtrace-library/src/androidTest/java/backtraceio/library/BacktraceCredentialsTest.java
```diff
@@ -76,4 +78,16 @@ public void createBacktraceCredentialsWithLegacyUriAndGetSubmissionToken() {
         BacktraceCredentials backtraceCredentials = new BacktraceCredentials(Uri.parse(legacyUrl));
         assertEquals(fakeToken, backtraceCredentials.getSubmissionToken());
     }
+
+    @Test
+    public void generateEmptyTokenForInvalidSubmissionUrl() {
+        BacktraceCredentials backtraceCredentials = new BacktraceCredentials("https://submit.backtrace.io/");
+        assertTrue(backtraceCredentials.getSubmissionToken() == null);
```
BartoszLitwiniuk, line 85: you can use assertNull here and below

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -567,15 +567,16 @@ private OnServerResponseEventListener getDatabaseCallback(final BacktraceDatabas
...
-                if (customCallback != null) {
-                    customCallback.onEvent(backtraceResult);
-                }
                 if (record != null) {
                     record.close();
                 }
                 if (backtraceResult != null && backtraceResult.status == BacktraceResultStatus.Ok) {
                     database.delete(record);
                 }
+
```
BartoszLitwiniuk, line 576: why is this part of this PR?

konraddysput: This was the initial ask of this pull request, but later I recognized it might be related to the invalid token. it would be great if we can leave it here, but if you prefer to have a separated diff for it - I'm ok with it and I will create a new one.

backtrace-library/src/androidTest/java/backtraceio/library/BacktraceCredentialsTest.java
konraddysput: Thank you - too many quick iterations on this.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: 👍

BartoszLitwiniuk (approved):
lgtm

melekr (approved):
lgtm

PR #96, opened 2023-05-23
Session id example

konraddysput:
# Why

This diff fixes session id in the sample application - by doing it in the wrong order, native errors might miss information about the current application session


BartoszLitwiniuk (approved):
lgtm

melekr (approved):
lgtm

PR #97, opened 2023-05-23
Handle invalid Crashpad handler path

konraddysput:
# Why

Backtrace-android must start a separate process in the system to handle crashes. This can be done via the crashpad handler path. By default we should always know the path to the handler, however, there are situations where libraries like crashpad handler are hidden or unavailable. Because of that, we have false positive information about the started process. 

This diff checks if the crashpad handler exists before we start the integration

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
+++ b/backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
@@ -177,12 +177,11 @@ public class BacktraceDatabase implements Database {
         }
         // Path to Crashpad native handler
         String handlerPath = _applicationContext.getApplicationInfo().nativeLibraryDir + _crashpadHandlerName;
-        File handlerPathFile = new File(handlerPath);
-        if (!handlerPathFile.exists()) {
+        if (!FileHelper.isFileExists(handlerPath)) {
             return false;
         }
 
-        // setup default native attribtues
+        // setup default native attributes
         BacktraceAttributes crashpadAttributes = new BacktraceAttributes(_applicationContext, client.attributes);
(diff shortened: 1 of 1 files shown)


BartoszLitwiniuk (approved):
lgtm

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -177,6 +177,10 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
         }
         // Path to Crashpad native handler
         String handlerPath = _applicationContext.getApplicationInfo().nativeLibraryDir + _crashpadHandlerName;
+        File handlerPathFile = new File(handlerPath);
+        if (!handlerPathFile.exists()) {
```
melekr, line 181: there is a method called `isFileExists` in class `FileHelper` that does exactly this

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -177,6 +177,10 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
...
+            return false;
+        }
 
         // setup default native attribtues
```
melekr, line 185: not yours but typo here

melekr (approved):
lgtm

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
konraddysput: good catch and also my typo! Thanks!

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
konraddysput: Cool, adjusted the code. thank you!

PR #104, opened 2023-09-07
Invalid intent flag

konraddysput:
# Why

We need to pass receiver flags from Android SDK 33+ to register a custom broadcast receiver. Otherwise, on Android SDK 34+, a missing flag will generate an error. 

This fixes a potential exception thrown in the Android SDK 34+ when no receiver flags are passed.

PR #105, opened 2023-09-14
Invalid intent flag #104

konraddysput:
# Why

We need to pass receiver flags from Android SDK 33+ to register a custom broadcast receiver. Otherwise, on Android SDK 34+, a missing flag will generate an error. 

This fixes a potential exception thrown in the Android SDK 34+ when no receiver flags are passed.


rick-bt (approved):
lgtm

PR #106, opened 2023-09-17
Fix delete same record multiple times

BartoszLitwiniuk:


backtrace-library/src/main/java/backtraceio/library/services/BacktraceDatabaseContext.java
```diff
@@ -175,13 +175,20 @@ public boolean delete(BacktraceDatabaseRecord record) {
...
 
                 databaseRecord.delete();
-                batchRetry.get(key).remove(databaseRecord);
+                try {
+                    records.remove(databaseRecord);
+                }
+                catch (Exception e) {
+                    BacktraceLogger.d(LOG_TAG, "Exception on removing record from db context: " + e.getMessage());
+                }
                 this.totalRecords--;
```
konraddysput, line 192: Are you should we should decrrase number of reports and database size? It looks like it is already handled by someone else

BartoszLitwiniuk: I was thinking about it, ok let's do this.

konraddysput (approved):

PR #108, opened 2023-10-27
Add session attribute and backtrace reporter name to every error report

konraddysput:
This diff adds `application.session` and `backtrace.agent` attributes to each report. 

The `application.session` is useful whether someone has metrics enabled or not. Using it, we can group all errors generated in a single application session. It's useful always to have it.

The `backtrace.agent` attribute is an attribute that we're adding to all our error-reporting libraries.


rick-bt (approved):

BartoszLitwiniuk (approved):
lgtm

PR #111, opened 2023-11-16
Remove check of throwable type in uncaughtException handler

BartoszLitwiniuk:
This pull request addresses a specific issue related to the uncaughtException handler. Currently, the code checks if the throwable object is of type Exception. However, this prevents the registration of errors such as OutOfMemoryError, which do not extend from the Exception class.

In this pull request, I am proposing the removal of the check for the specific type Exception to ensure that errors like OutOfMemoryError are appropriately captured and handled.

Please review the changes at your earliest convenience. I believe this modification will improve the robustness of our error handling mechanism. This pull requests contains also unit tests for both cases: exception and error.

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
+++ b/backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
@@ -570,7 +570,8 @@ public class BacktraceBase implements Client {
                 if (record != null) {
                     record.close();
                 }
-                if (backtraceResult != null && backtraceResult.status == BacktraceResultStatus.Ok) {
+
+                if (backtraceResult != null && record != null && backtraceResult.status == BacktraceResultStatus.Ok) {
                     database.delete(record);
                 }
 
--- a/backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
+++ b/backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
@@ -70,9 +70,8 @@ public class BacktraceExceptionHandler implements Thread.UncaughtExceptionHandle
             return (Exception) throwable;
         }
 
-        Exception exception = new Exception(throwable);
+        Exception exception = new Exception(throwable.toString(), throwable);
         exception.setStackTrace(throwable.getStackTrace());
-
         return exception;
     }
 
--- a/backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
+++ b/backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
@@ -53,23 +53,18 @@ public class BacktraceExceptionHandler implements Thread.UncaughtExceptionHandle
     public void uncaughtException(final Thread thread, final Throwable throwable) {
         OnServerResponseEventListener callback = getCallbackToDefaultHandler(thread, throwable);
 
-        if (throwable instanceof Exception) {
-            BacktraceLogger.e(LOG_TAG, "Sending uncaught exception to Backtrace API", throwable);
-            BacktraceReport report = new BacktraceReport((Exception) throwable, BacktraceExceptionHandler.customAttributes);
-            report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
-            this.client.send(report, callback);
-            BacktraceLogger.d(LOG_TAG, "Uncaught exception sent to Backtrace API");
-        }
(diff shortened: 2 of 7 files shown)


backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -570,7 +570,8 @@ public void onEvent(BacktraceResult backtraceResult) {
                 if (record != null) {
                     record.close();
                 }
-                if (backtraceResult != null && backtraceResult.status == BacktraceResultStatus.Ok) {
+
+                if (backtraceResult != null && record != null && backtraceResult.status == BacktraceResultStatus.Ok) {
                     database.delete(record);
```
konraddysput, line 575: I think the record check should be in the database code rather than in the client. If we want to have such a check, can we move it to the database instead?

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -53,23 +51,28 @@ public static void enable(BacktraceClient client) {
...
+        report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
+        this.client.send(report, callback);
+        BacktraceLogger.d(LOG_TAG, "Uncaught exception sent to Backtrace API");
+
         try {
             signal.await();
         } catch (Exception ex) {
             BacktraceLogger.e(LOG_TAG, "Exception during waiting for response", ex);
         }
+        BacktraceLogger.d(LOG_TAG, "Default uncaught exception handler");
```
konraddysput, line 65: I think default unhandled exception handler is being executed on line 61 - isn't it?

konraddysput (approved):

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -53,23 +51,28 @@ public static void enable(BacktraceClient client) {
...
-            BacktraceReport report = new BacktraceReport((Exception) throwable, BacktraceExceptionHandler.customAttributes);
-            report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
-            this.client.send(report, callback);
-            BacktraceLogger.d(LOG_TAG, "Uncaught exception sent to Backtrace API");
-        }
-        BacktraceLogger.d(LOG_TAG, "Default uncaught exception handler");
+        BacktraceLogger.e(LOG_TAG, "Sending uncaught exception to Backtrace API", throwable);
+        BacktraceReport report = new BacktraceReport(this.getCausedException(throwable), BacktraceExceptionHandler.customAttributes);
+        report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
+        this.client.send(report, callback);
```
konraddysput, line 57: since we're creating and passing a report to the client, can we also pass a throwable type as classifier? Can we override classifiers in the report? Now, we will pass a generic type of classifier which isn't great.

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: it's extra check otherwise we may have weird nulls inside database context

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
BartoszLitwiniuk: Fixed

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
BartoszLitwiniuk: Good point, fixed

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
konraddysput: Yeah but it would be better if we put it in the database context entry point instead

konraddysput (approved):

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
BartoszLitwiniuk: I've added extra null check in BacktraceDatabase, now we have it there, here and in BacktraceDatabaseContext

PR #112, opened 2023-11-24
Handle potential serialization exceptions during custom exception serialization

konraddysput:
# Why

This diff allows us to avoid potential exception serialization issues during the report/data serialization in the database/sender. 

Goal:
- serialize only known exception properties
- generate the same information we have now (no change for user)


BartoszLitwiniuk (approved):
lgtm

PR #117, opened 2024-02-07
New client attribute API

konraddysput:
# Why

Our Backtrace client doesn't have any programmatic interface to add attributes. We allow our customers to use direct attributes map, however, we don't know when it's happening and because of this we cannot update native integration attributes. All attributes added dynamically by the user during the application lifecycle should be visible in the native integration.

This pull request adds a new API that allows adding an attribute key pair/ multiple attributes to the Backtrace Client and updates managed/native integration attributes 

ref: BT-2255

### DIFF
--- a/backtrace-library/src/androidTest/java/backtraceio/library/database/BacktraceDatabaseTest.java
+++ b/backtrace-library/src/androidTest/java/backtraceio/library/database/BacktraceDatabaseTest.java
@@ -52,6 +52,10 @@ public class BacktraceDatabaseTest {
         this.database.clear();
     }
 
+    @Test
+    public void shouldNotCrashWhenNativeIntegrationIsNotEnabled() {
+        assertEquals(false, this.database.addNativeAttribute("key", "value"));
+    }
 
     @Test
     public void isDatabaseEmpty() {
--- a/backtrace-library/src/androidTest/java/backtraceio/library/database/BacktraceDatabaseTest.java
+++ b/backtrace-library/src/androidTest/java/backtraceio/library/database/BacktraceDatabaseTest.java
@@ -3,6 +3,8 @@ package backtraceio.library.database;
 import static org.junit.Assert.assertEquals;
 
 import android.content.Context;
+import android.os.FileUtils;
+
 import androidx.test.platform.app.InstrumentationRegistry;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
@@ -14,7 +16,9 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import java.io.File;
+import java.io.FileFilter;
 import java.io.IOException;
+import java.nio.file.FileSystems;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -36,6 +40,8 @@ public class BacktraceDatabaseTest {
     private Context context;
     private String dbPath;
     private BacktraceDatabase database;
+    private BacktraceClient client;
+    private BacktraceCredentials credentials;
     private final String testMessage = "Example test string";
 
     @Before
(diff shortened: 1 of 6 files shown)


backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -296,6 +296,35 @@ public Map<String, Object> getAttributes() {
...
+    }
+    /**
+     * Adds new attributes to Backtrace Client. If the native integration is enabled, adds attributes
+     * to the native report attributes if:
+     *  - the value exists (is not a null)
+     *  - is not an object (the attribute value is primitive type like String, or Int)
+     * @param attributes Map of attributes
+     */
+    public void addAttribute(Map<String, Object> attributes) {
+        for (Map.Entry<String, Object> entry : attributes.entrySet()) {
```
BartoszLitwiniuk, line 323: if attributes will be null it will crash

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -296,6 +296,35 @@ public Map<String, Object> getAttributes() {
...
+    /**
+     * Adds a new attribute to Backtrace Client. If the native integration is enabled, adds the attribute
+     * to the native report attributes if:
+     *  - the value exists (is not a null)
+     *  - is not an object (the attribute value is primitive type like String, or Int)
+     * @param key attribute name
+     * @param value attribute value.
+     */
+    public void addAttribute(String key, Object value) {
+        attributes.put(key, value);
```
BartoszLitwiniuk, line 308: what if value will be null

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -50,6 +51,7 @@ public class BacktraceDatabase implements Database {
     private boolean _enable = false;
     private Breadcrumbs breadcrumbs;
 
+    private boolean _enabledNativeIntegration = false;
```
BartoszLitwiniuk, line 54: name can be without _

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -167,7 +169,7 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
     public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials credentials,
                                           boolean enableClientSideUnwinding, UnwindingMode unwindingMode) {
         // avoid initialization when database doesn't exist
-        if (getSettings() == null) {
+        if (_enable == false || getSettings() == null) {
```
BartoszLitwiniuk, line 172: `if (!_enable || getSettings() == null) {`

backtrace-library/src/androidTest/java/backtraceio/library/database/BacktraceDatabaseTest.java
```diff
@@ -52,6 +52,10 @@ public void after() {
         this.database.clear();
     }
 
+    @Test
+    public void shouldNotCrashWhenNativeIntegrationIsNotEnabled() {
```
BartoszLitwiniuk, line 56: please add test which adding null as key and null as value and check behaviour

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
konraddysput: In following the pattern that we already have there

backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
konraddysput: It's fine if the value is null. We want to store this information. We cannot pass it to the native

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
konraddysput: I will stick to the current implementation

BartoszLitwiniuk (approved):
lgtm

PR #121, opened 2024-04-10
Fix unsupported metrics server

BartoszLitwiniuk:


backtrace-library/src/main/java/backtraceio/library/base/BacktraceBase.java
```diff
@@ -233,12 +244,12 @@ public BacktraceBase(Context context, BacktraceCredentials credentials, Database
...
         this.credentials = credentials;
-        this.attributes = attributes != null ? attributes : new HashMap<String, Object>();
-        this.attachments = attachments != null ? attachments : new ArrayList<String>();
+        this.attributes = attributes != null ? attributes : new HashMap<>();
+        this.attachments = attachments != null ? attachments : new ArrayList<>();
         this.database = database != null ? database : new BacktraceDatabase();
         this.setBacktraceApi(new BacktraceApi(credentials));
         this.database.start();
-        this.metrics = new BacktraceMetrics(context, attributes, backtraceApi, credentials);
+        this.metricsController = new BacktraceMetricsController(context, attributes, backtraceApi, credentials);
```
konraddysput, line 252: can we please keep current implementation and avoid additional abstraction complexity? We should try to simplify our library for every use case and follow what we already have. The expectation now is to fail on startup, rather than fail silently

PR #122, opened 2024-04-16
Metrics fix unsupported server

BartoszLitwiniuk:


backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
```diff
@@ -155,11 +158,17 @@ public void enable(BacktraceMetricsSettings settings) {
     }
 
     public void enable(BacktraceMetricsSettings settings, String uniqueEventName) {
+        if (!settings.isBacktraceServer()) {
```
konraddysput, line 161: lets also check if metrics are already enabled. Lets avoid a situation when we enable metrics twice.

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
```diff
@@ -360,8 +378,8 @@ private boolean shouldProcessEvent(String name) {
         return true;
     }
 
-    protected Map<String, Object> createLocalAttributes(Map<String, Object> attributes) {
-        Map<String, Object> localAttributes = new HashMap<String, Object>();
+    Map<String, Object> createLocalAttributes(Map<String, Object> attributes) {
```
konraddysput, line 381: can we keep it obvious? I think its fine to leave protected modified here. I know its default behavior but we keep accessor definition everywhere.

konraddysput (changes requested):
Lets throw an error on double initialization - or lets skip double initialization

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
BartoszLitwiniuk: fixed

konraddysput (approved):

PR #128, opened 2024-05-28
Crashpad - replace crashpad_handler executable with JavaCrashHandler

konraddysput:
# Why

This diff replaces the old strategy of using crashpad_handler executable with the new JavaCrashHandler setup. Due to updates to AGP, by default native libraries are not extracted. Because of this, our customers need to use useLegacyPackaging flag to use our NDK support.

To avoid using legacy packaging flags, we wanted to change the way how Crashpad works. Instead of using crashpad_handler executable, this pull request exposes the `BacktraceCrashHandler` class that will be spawned by app_process when the crash occurs. The class is a wrapper around Crashpad dumping API, that allows us to generate a minidump and send it the same way like crashpad handler would do.

To avoid too big scope of the change, in future pull requests, we want to:
- fix a problem with invalid guid override API behavior
- expose JavaCrashHandler under a different method in the crashapd-backend.


backtrace-library/build.gradle
```diff
@@ -12,12 +12,10 @@ androidGitVersion {
 android {
 
     compileSdkVersion 33
-
+    namespace "backtraceio.library"
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 33
-        versionName androidGitVersion.name()
```
BartoszLitwiniuk, line 19: why?

backtrace-library/publish.gradle
```diff
@@ -111,12 +111,10 @@ afterEvaluate { project ->
         }
 
         task androidJavadocsJar(type: Jar, dependsOn: androidJavadocs) {
-            classifier = 'javadoc'
```
BartoszLitwiniuk, line 114: why are you removing that and below

backtrace-library/src/androidTest/java/backtraceio/library/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="backtraceio.library">
```
BartoszLitwiniuk, line 2: why such change

example-app/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
-        android:extractNativeLibs="true">
```
BartoszLitwiniuk, line 15: why?

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+
+        Log.i(LOG_TAG, "Successfully ran crash handler code.");
+    }
+
+    public String getClassPath() {
+        return BacktraceCrashHandler.class.getCanonicalName();
+    }
+
+    public String[] setCrashHandlerEnvironmentVariables(ApplicationInfo applicationInfo) {
+        List<String> environmentVariables = new ArrayList<>();
```
BartoszLitwiniuk, line 48: can be final

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+public class BacktraceCrashHandler {
+
+    public static native boolean handleCrash(String[] args);
+
+    private static final String LOG_TAG = BacktraceCrashHandler.class.getSimpleName();
+    private static final String BACKTRACE_CRASH_HANDLER = "BACKTRACE_CRASH_HANDLER";
+
+    public static void main(String[] args) {
+        Log.i(LOG_TAG, "Starting Backtrace crash handler code.");
+        Map<String, String> environmentVariables = System.getenv();
```
BartoszLitwiniuk, line 21: can be final

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        environmentVariables.add("CLASSPATH=" + applicationInfo.sourceDir);
+        environmentVariables.add(BACKTRACE_CRASH_HANDLER + "=" + applicationInfo.sourceDir + "!/lib/arm64-v8a/libbacktrace-native.so");
+        environmentVariables.add("LD_LIBRARY_PATH=" + TextUtils.join(File.pathSeparator, new String[]{
+                applicationInfo.nativeLibraryDir,
+                new File(applicationInfo.nativeLibraryDir).getParentFile().getPath(),
+                System.getProperty("java.library.path"),
+                "/data/local"}));
+
+        environmentVariables.add("ANDROID_DATA=/data");
+
```
BartoszLitwiniuk, line 64: remove empty line

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        for (Map.Entry<String, String> variable :
+                System.getenv().entrySet()) {
+            environmentVariables.add(variable.getKey() + "=" + variable.getValue());
+        }
+
+        environmentVariables.add("CLASSPATH=" + applicationInfo.sourceDir);
+        environmentVariables.add(BACKTRACE_CRASH_HANDLER + "=" + applicationInfo.sourceDir + "!/lib/arm64-v8a/libbacktrace-native.so");
+        environmentVariables.add("LD_LIBRARY_PATH=" + TextUtils.join(File.pathSeparator, new String[]{
+                applicationInfo.nativeLibraryDir,
+                new File(applicationInfo.nativeLibraryDir).getParentFile().getPath(),
```
BartoszLitwiniuk, line 59: please extract to new method and call it here
```
new File(applicationInfo.nativeLibraryDir).getParentFile().getPath()
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Mar 29 15:48:21 CEST 2023
+#Mon May 20 12:47:38 CEST 2024
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
```
BartoszLitwiniuk, line 4: why not latest?

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -176,11 +182,15 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
...
             return false;
         }
-        // Path to Crashpad native handler
-        String handlerPath = _applicationContext.getApplicationInfo().nativeLibraryDir + _crashpadHandlerName;
-        if (!FileHelper.isFileExists(handlerPath)) {
-            return false;
-        }
+
+        // Create the crashpad directory if it doesn't exist
+        String databasePath = getSettings().getDatabasePath() + _crashpadDatabasePathPrefix;
```
BartoszLitwiniuk, line 187: extract this block of code to method `createCrashpadDirectory`

btw what if this directory already exists?

backtrace-library/src/main/java/backtraceio/library/BacktraceDatabase.java
```diff
@@ -176,11 +182,15 @@ public Boolean setupNativeIntegration(BacktraceBase client, BacktraceCredentials
...
+        File crashHandlerDir = new File(databasePath);
+        crashHandlerDir.mkdir();
+
+        String classPath = this.crashHandler.getClassPath();
```
BartoszLitwiniuk, line 191: extract this method to separate function

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -122,25 +124,54 @@ bool InitializeCrashpad(jstring url,
...
+                                                  arguments);
 
     env->ReleaseStringUTFChars(url, backtraceUrl);
     env->ReleaseStringUTFChars(handler_path, handlerPath);
     env->ReleaseStringUTFChars(database_path, filePath);
 
-    if (enableClientSideUnwinding) {
-        SetCrashpadHandlerForClientSideUnwinding();
-    }
+//    if (enableClientSideUnwinding) {
```
BartoszLitwiniuk, line 154: do we need it?

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+import android.content.pm.ApplicationInfo;
+import android.text.TextUtils;
+import android.util.Log;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class BacktraceCrashHandler {
```
BartoszLitwiniuk, line 12: Please add unit tests for methods in this file

backtrace-library/publish.gradle
konraddysput: This is required by gradle upgrade to version 8.2.0

backtrace-library/src/androidTest/java/backtraceio/library/AndroidManifest.xml
konraddysput: This is required by gradle upgrade to version 8.2.0

example-app/src/main/AndroidManifest.xml
konraddysput: This is required by gradle upgrade to version 8.2.0

gradle/wrapper/gradle-wrapper.properties
konraddysput: Bumping higher requires more changes and extends the scope. I would prefer to not extend the scope too much since we're already introducing a lot of changes

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
konraddysput: yes - for now we need to keep it.

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+    public String[] setCrashHandlerEnvironmentVariables(ApplicationInfo applicationInfo) {
+        List<String> environmentVariables = new ArrayList<>();
+
+        for (Map.Entry<String, String> variable :
+                System.getenv().entrySet()) {
+            environmentVariables.add(variable.getKey() + "=" + variable.getValue());
+        }
+
+        environmentVariables.add("CLASSPATH=" + applicationInfo.sourceDir);
+        environmentVariables.add(BACKTRACE_CRASH_HANDLER + "=" + applicationInfo.sourceDir + "!/lib/arm64-v8a/libbacktrace-native.so");
```
konraddysput, line 56: hardcoded arm64 path that needs to change

backtrace-library/build.gradle
konraddysput: AGP in the higher version removed it for me. I bring it back to limit the scope.

backtrace-library/src/main/java/backtraceio/library/services/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        Log.i(LOG_TAG, "Successfully ran crash handler code.");
+    }
+
+    public String getClassPath() {
+        return BacktraceCrashHandler.class.getCanonicalName();
+    }
+
+    public String[] setCrashHandlerEnvironmentVariables(ApplicationInfo applicationInfo) {
```
gmegidish-saucelabs, line 47: Naming is incorrect imho. You're getting environment variables, not set them.

gmegidish-saucelabs: Would you prefer to change this to Map instead of String[] for the return value? then it would be easier to fetch values for keys through JNI, and you won't have to format strings by adding "=" all the time. Wdyt?

backtrace-library/build.gradle
```diff
@@ -77,6 +80,9 @@ dependencies {
     androidTestImplementation 'androidx.test:rules:1.5.0'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
+    androidTestImplementation 'org.mockito:mockito-core:2.24.5'
```
BartoszLitwiniuk, line 83: 2.28.2 is latest https://mvnrepository.com/artifact/org.mockito/mockito-core/2.28.2

BartoszLitwiniuk (approved):
lgtm

PR #131, opened 2024-06-03
Bring back support for crash_handler.so for backward compatibility

konraddysput:
# Why

This pull request returns support for libcrashpad_handler executable available in the apk. WIth this change, our NDK interface doesn't introduce any breaking change for other our libraries

### DIFF
--- a/backtrace-library/src/main/cpp/backends/backend.cpp
+++ b/backtrace-library/src/main/cpp/backends/backend.cpp
@@ -11,7 +11,7 @@
 #endif
 
 extern std::atomic_bool initialized;
-
+static std::once_flag initialize_flag;
 extern "C" {
 bool Initialize(jstring url,
                 jstring database_path,
@@ -20,9 +20,8 @@ bool Initialize(jstring url,
                 jobjectArray attributeValues,
                 jobjectArray attachmentPaths,
                 jboolean enableClientSideUnwinding,
-                jint unwindingMode,
-                jobjectArray environmentVariables) {
-    static std::once_flag initialize_flag;
+                jint unwindingMode) {
+
 
     std::call_once(initialize_flag, [&] {
 #ifdef CRASHPAD_BACKEND
@@ -30,7 +29,7 @@ bool Initialize(jstring url,
                                          database_path, handler_path,
                                          attributeKeys, attributeValues,
                                          attachmentPaths, enableClientSideUnwinding,
-                                         unwindingMode, environmentVariables);
+                                         unwindingMode);
 #elif BREAKPAD_BACKEND
         initialized = InitializeBreakpad(url,
                                          database_path, handler_path,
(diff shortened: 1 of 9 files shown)


BartoszLitwiniuk (approved):
lgtm

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -12,18 +12,98 @@ extern std::atomic_bool disabled;
...
+        std::string attachmentBaseName = basename(convertedAttachmentPath);
+
+        std::string attachmentArgumentString("--attachment=");
+        attachmentArgumentString += convertedAttachmentPath;
+        arguments.push_back(attachmentArgumentString);
+
+        env->ReleaseStringUTFChars(jstringAttachmentPath, convertedAttachmentPath);
+        env->DeleteLocalRef(jstringAttachmentPath);
+    }
+
```
gm4sl, line 45: (nit) extra newlines in diff on Line 46 and 48.

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -256,7 +353,8 @@ void ReEnableCrashpad() {
     // Re-enable uploads if disabled
     if (disabled) {
         if (database == nullptr) {
-            __android_log_print(ANDROID_LOG_ERROR, "Backtrace-Android", "Crashpad database is null, this should not happen");
```
gm4sl: Seems like these hunks should be moved back to the NFC formatting commit. It would make it easier to determine what actually needs to be reviewed.

backtrace-library/src/main/cpp/backends/backend.cpp
gm4sl, line 1: Is there a reason the crash handler executable path was removed in the first place? If another PR is still open can we update it to leave the existing path alone? If not, this PR is for bringing back the old crash handler executable path but adds, what seems to be, a lot of new code (like generateInitializationArguments). Is this new code that needs to be reviewed or is it previous code that was deleted and is being revived here? It is hard to tell.

gm4sl (changes requested):

backtrace-library/src/main/cpp/backends/backend.cpp
konraddysput: Update old pull request - we talked about it during our session. The reason why I wanted to do it in the separated pull request is to make sure you can easily track those changes. I felt it could be confusing for you if I changed everything after reviewing code by other engineers.

The old path and the new one has a lot of the same code. I exposed this code into function so both flows can use it.

This code change is about making those two parts working together.

backtrace-library/src/main/cpp/backends/backend.cpp
```diff
@@ -11,7 +11,7 @@
 #endif
 
 extern std::atomic_bool initialized;
-
+static std::once_flag initialize_flag;
```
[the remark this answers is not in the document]
konraddysput: We have two way to initialize crashpad - via crash handler executabke and Java class. Our assumption is, we cannot have two crash reporting solution working together. With that being said, we need to verify if we already initialized the client. That's why it exists in both places and for the same variable

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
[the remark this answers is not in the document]
konraddysput: When I checked code I saw there is no reason to parse attributes if later the code can fail due to invalid database condition. That's why I moved my attributes generation to a different place. It can be in the old place but the new order doesn't do something we don't need to do in the fail path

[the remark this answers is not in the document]
gm4sl: Makes sense. Can we include that reasoning in the log?

backtrace-library/src/main/cpp/backends/backend.cpp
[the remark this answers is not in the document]
gm4sl: That's what the atomic `initialized` is for, I think. It is probably not that big of a deal though.

backtrace-library/src/main/cpp/backends/backend.cpp
gm4sl: I'm not sure I totally understand but if it was done intentionally that is fine. It is, for me, harder to track what was change in this PR.

PR #135, opened 2024-06-11
Include missing aar in the final library 

konraddysput:
# Why

Update to the latest AGP required some changes in the publish/build gradle setting. We don't need to use some options, since they're already available in the platform. With this update, we should be able to ship our libraries with aar.


BartoszLitwiniuk (approved):
lgtm

PR #139, opened 2024-06-14
Use extracted libraries when they're available

konraddysput:
# Why

We recognized when native libraries are extracted, we cannot use the path to ndk libraries inside the apk that we were using. By checking if the library is already extracted we can check which one native library should handle dump request code.


BartoszLitwiniuk (approved):
lgtm

PR #140, opened 2024-07-22
Use getter getReport instead of report field

BartoszLitwiniuk:
Change to use backtrace data getter for report instead of field directly.

Motivation:
Using getters instead of directly accessing fields in Java ensures encapsulation, allows for data validation, and provides control over how fields are accessed and modified. This practice enhances code flexibility, supports design patterns, and enables easier debugging and maintenance by allowing internal implementation changes without affecting external code.

What is more in the future we should consider moving or removing BacktraceReport field from BacktraceData because we are not serializing and using that in proper way. When this change will be apply we can easily control that without breaking API.


konraddysput:
What is the motivation behind this change?

BartoszLitwiniuk:
I've updated PR description

konraddysput (approved):

PR #142, opened 2024-08-05
Update sauce test suite

konraddysput:
# Why

This pull request should fix the problem with not available devices in Sauce for really old version of Android.

ref: BT-2909

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,7 @@ jobs:
           # oldest Android version
           saucectl run espresso -c "" --name "From Github Actions" --app example-app-debug.apk --testApp ${{ matrix.test-apk }}  --emulator name="Android GoogleApi Emulator,platformVersion=8.0" --region us-west-1
           # oldest real device (aiming for armv7 / 32bit)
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -106,9 +106,9 @@ jobs:
           # newest Android version
           saucectl run espresso -c "" --name "From Github Actions" --app example-app-debug.apk --testApp ${{ matrix.test-apk }}  --emulator name="Android GoogleApi Emulator,platformVersion=12.0" --region us-west-1
           # oldest Android version
-          saucectl run espresso -c "" --name "From Github Actions" --app example-app-debug.apk --testApp ${{ matrix.test-apk }}  --emulator name="Android GoogleApi Emulator,platformVersion=6.0" --region us-west-1
+          saucectl run espresso -c "" --name "From Github Actions" --app example-app-debug.apk --testApp ${{ matrix.test-apk }}  --emulator name="Android GoogleApi Emulator,platformVersion=8.0" --region us-west-1
```
rick-bt, line 109: I think the emulator is ok with 6.0? I only hit an error with physical devices.

rick-bt: Minimum supported device in VDC is Android 5.

konraddysput: I think it can be dropped any time since we want to remove legacy support for old os

rick-bt (approved):

PR #143, opened 2024-08-12
Run tests against public cloud

konraddysput:
# Why

We have a lot of devices in the public cloud. We don't need to use private devices to run tests. If the device is for some kind of reason offline, then our CICD pipeline can still be executed against any of the multiple devices available in the cloud

ref: BT-2909


rick-bt (approved):
Approved. Can we remove the second test? Or bump the version up?

rick-bt (approved):

PR #147, opened 2024-08-21
Feature/custom logger

BartoszLitwiniuk:
## Summary
This PR introduces support for replacing the internal logger of the Backtrace Android SDK with a custom implementation. This allows developers to integrate their own logging systems for better control and consistency.

## Key Changes
- Custom Logger Support: Added a new interface for injecting custom loggers into the SDK.
- Backward Compatibility: Retained the default logger for users who do not opt for customization.
- Documentation: Updated the README with instructions on implementing and configuring a custom logger.

## Benefits
- Flexibility: Align SDK logging with your application's logging system.
- Customization: Control log levels, formats, and outputs.

### DIFF
--- a/backtrace-library/build.gradle
+++ b/backtrace-library/build.gradle
@@ -77,18 +77,18 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation 'com.google.code.gson:gson:2.10.1'
+    implementation 'com.google.code.gson:gson:2.11.0'
     implementation 'androidx.appcompat:appcompat:1.6.1'
     implementation 'com.squareup:tape:1.2.3'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.2.0'
     testImplementation "org.mockito:mockito-android:2.28.2"
     androidTestImplementation 'net.jodah:concurrentunit:0.4.4'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.5'
-    androidTestImplementation 'androidx.test:rules:1.5.0'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.5'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
-    androidTestImplementation 'org.mockito:mockito-core:2.28.2'
+    androidTestImplementation 'androidx.test.ext:junit:1.2.1'
+    androidTestImplementation 'androidx.test:rules:1.6.1'
+    androidTestImplementation 'androidx.test.ext:junit:1.2.1'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.6.1'
+    androidTestImplementation 'org.mockito:mockito-core:5.2.0'
     androidTestImplementation "org.mockito:mockito-android:2.28.2"
 }
 
--- a/backtrace-library/build.gradle
+++ b/backtrace-library/build.gradle
@@ -81,8 +81,8 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.6.1'
     implementation 'com.squareup:tape:1.2.3'
     testImplementation 'junit:junit:4.13.2'
-    testImplementation 'org.mockito:mockito-core:5.2.0'
-    testImplementation "org.mockito:mockito-android:2.28.2"
+    testImplementation "org.mockito:mockito-core:5.12.0"
+    androidTestImplementation "org.mockito:mockito-android:5.12.0"
     androidTestImplementation 'net.jodah:concurrentunit:0.4.4'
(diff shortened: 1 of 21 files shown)


backtrace-library/build.gradle
```diff
@@ -81,6 +81,8 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.6.1'
```
BartoszLitwiniuk, line 81: TODO: update README how to use it

BartoszLitwiniuk: it will be moved to internal saucelabs doc

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
```diff
@@ -1,25 +1,25 @@
...
-     */
-    private static LogLevel logLevel = LogLevel.OFF;
+    private static Logger logger = new BacktraceLogLogger();
 
-    /**
-     * set logging level from which all messages should be logged to the console
-     *
-     * @param level login level
-     */
-    public static void setLevel(LogLevel level) {
```
konraddysput, line 21: Unfortunately we should keep it. Otherwise this will be a breaking change. I know some customers are using it

konraddysput (changes requested):
We cannot introduce any breaking change. Can we recover setLevel?

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogLogger.java
```diff
@@ -0,0 +1,95 @@
...
+     * Level from which all information is logged
+     */
+    private LogLevel logLevel;
+
+    public BacktraceLogLogger() {
+        this(LogLevel.OFF);
+    }
+
+    public BacktraceLogLogger(LogLevel logLevel) {
+        this.logLevel = logLevel;
```
konraddysput, line 26: In each logging function you're executing:
`this.logLevel.ordinal()` 

since this is critical path that can be executed quite often, maybe we should set `this.logLevel =logLevel.ordinal();` instead?

BartoszLitwiniuk: good point, improved

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
```diff
@@ -1,25 +1,25 @@
...
-        BacktraceLogger.logLevel = level;
+    public static void setLogger(Logger logger) {
```
melekr, line 21: Can we add some kind of integrity or null check?  in case someone hands us a lemon 😄

melekr (approved):

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
BartoszLitwiniuk: Good point, added with deprecated annotation

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
BartoszLitwiniuk: thanks, I've added null check

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
```diff
@@ -71,13 +71,20 @@ public static int e(String tag, String message) {
...
+     * @param level login level (debug, warn, error, off)
+     *
+     * @deprecated setting the logging level should be done directly in the passed custom logger
+     * implementation so as not to depend on the internal implementation of Backtrace internal
+     * LogLevel types.
+     *
+     */
+    @Deprecated
+    public static void setLevel(@NotNull LogLevel level) {
+        logger.setLevel(level);
```
konraddysput, line 88: lets think about this twice

Since this is the interface that we want to deprecate anyway, I think the logger interface will have some deprecated API as well. 

My suggestion is: 

- can we remove setLevel method from the logger definition? 
- since this is deprecated and we want to remove it - maybe we can use log level in this class and then pass data to Logger? 

With this approach we have less fields to deprecate and less changes the customer needs to make

konraddysput (commented):
Adding one more important suggestion to the pull request. Lets resolve this conversation and I'm ready to accept the PR

BartoszLitwiniuk:
PR with docs https://github.com/saucelabs/sauce-docs/pull/2925/files.

backtrace-library/src/main/java/backtraceio/library/logger/BacktraceLogger.java
BartoszLitwiniuk: please review one more time

konraddysput (approved):

PR #151, opened 2024-09-25
Logging debug time of breadcrumbs init and setup native integration

BartoszLitwiniuk:
Log information about how long took breadcrumbs init and setup native integration for debug purposes.


konraddysput (approved):
I think we should also check metrics initialization code

PR #157, opened 2024-11-05
Metrics performance improvements 

konraddysput:
# Why

Attribute generations .... cost us time. We generated them multiple times during metrics setup AND metrics payload generation. In addition to that we had a situation where we generated attributes twice during single object creation (in the constructor to get an application name and later to get all attributes for payload)

This pull request significantly improves the performance of the metrics system - my internal results give me 2x time reduction. Other parts of our system can also benefit from it - everything that uses our attribute system should retrieve results faster.

ref: BT-2660


backtrace-library/src/androidTest/java/backtraceio/library/BacktraceClientUniqueEventTest.java
```diff
@@ -400,48 +400,4 @@ public void uniqueEventUpdateAttributes() {
 
         assertEquals(expectedValue, backtraceClient.metrics.getUniqueEvents().getLast().getAttributes().get(expectedKey));
     }
-
-    @Test
-    public void uniqueEventEmptyAttributeValueShouldNotOverridePreviousValueOnUpdate() {
```
konraddysput, line 405: those tests don't make sense. If the user changes the attribute, the new attribute value should be send to metrics. It can generate false positive data if we won't change the value

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
```diff
@@ -0,0 +1,74 @@
+package backtraceio.library.common;
+
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+
+import backtraceio.library.logger.BacktraceLogger;
+
+public class ApplicationHelper {
```
BartoszLitwiniuk, line 9: WDYT about naming that e.g. `AppInfoCache`/`ApplicationMetadata`/`CachedAppInfo`

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
```diff
@@ -0,0 +1,74 @@
+package backtraceio.library.common;
```
BartoszLitwiniuk, line 1: No unit tests for this file

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -36,6 +36,11 @@
 public class DeviceAttributesHelper {
     private final Context context;
 
+    /*
+     * Current Device id
+     */
+    private static String guid;
```
BartoszLitwiniuk, line 42: in Java it should be `uuid`

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
```diff
@@ -0,0 +1,74 @@
...
+     * Cached package name
+     */
+    private static String packageName;
+
+    /**
+     * Retrieves application name from context. The name will be cached over checks
+     * @param context application context
+     * @return application name
+     */
+    public static String getApplicationName(Context context) {
```
BartoszLitwiniuk, line 31: All of your static methods requires Context so why not pass it through constructor?

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
```diff
@@ -0,0 +1,40 @@
+package backtraceio.library.models.json;
```
BartoszLitwiniuk, line 1: Reformat this file

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+        HashMap<String, Object> reportAnnotations = new HashMap<>();
+
+        if(attributes == null) {
+            return new Tuple<>(reportAttributes, reportAnnotations);
+        }
+
+        for (Map.Entry<String, Object> entry : attributes.entrySet()) {
+            Object value = entry.getValue();
+            if (value == null) {
+                reportAttributes.put(entry.getKey(), "null");
```
BartoszLitwiniuk, line 26: why we need to have string 'null' instead of normal null?

BartoszLitwiniuk:
to be honest I really don't like introducing tuples because of 
- lack of Readability and self-Documentation
- weak type safety
- java's type system is designed for strongly-typed, structured objects rather than lightweight, anonymous data structures like tuples

It's easier just to wrap two fields with small object and pass it together if they are logically combined, if not then maybe code is bad designed.

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Retrieves application version from the context. If the version name is not defined, the version code will be used instead.
+     * @param context application context
+     * @return current application version.
+     */
+    public static String getApplicationVersion(Context context) {
+        if(!BacktraceStringHelper.isNullOrEmpty(applicationVersion)) {
+            return applicationVersion;
+        }
+        try {
```
melekr, line 49: if we want to avoid a try/catch, we can also use Gradle's BuildConfig.VERSION_CODE/BuildConfig.VERSION_NAME for applicationVersion.

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
```diff
@@ -211,41 +203,12 @@ private void convertReportAttributes(BacktraceReport report) {
...
-            }
-        }
-    }
-
-    public String getApplicationName() {
-        return this.context.getApplicationInfo().loadLabel(this.context
-                .getPackageManager()).toString();
+    private void convertAttributes(Map<String, Object> clientAttributes) {
+        Tuple<Map<String, String>, Map<String, Object>> reportData = ReportDataBuilder.getReportAttribues(clientAttributes);
+        this.attributes.putAll(reportData.first);
```
melekr, line 208: a null check here would be great

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Divide custom user attributes into primitive and complex attributes and add to this object
+     *
+     * @param attributes client's attributes
+     */
+    public static Tuple<Map<String, String>, Map<String, Object>> getReportAttribues(Map<String, Object> attributes) {
+        HashMap<String, String> reportAttributes = new HashMap<>();
+        HashMap<String, Object> reportAnnotations = new HashMap<>();
+
+        if(attributes == null) {
```
melekr, line 19: missing a space here

backtrace-library/src/main/java/backtraceio/library/models/metrics/Event.java
```diff
@@ -12,8 +12,11 @@ public abstract class Event {
     @SerializedName("timestamp")
     protected long timestamp;
 
+    /**
+     * The event should always have attributes. Event without attributes won't be processed
+     */
     @SerializedName("attributes")
-    protected Map<String, Object> attributes;
+    protected Map<String, String> attributes = new HashMap<>();
```
melekr, line 19: We can also use synchronized for thread safety : 
`protected final Map<String, String> attributes = Collections.synchronizedMap(new HashMap<>());`

and access it like:

`public Map<String, String> getAttributes() {
        synchronized (attributes) {
            return new HashMap<>(this.attributes);  // Return a copy to prevent modification
        }
    }`

`protected void addAttributes(Map<String, String> attributes) {
        if (attributes == null || attributes.isEmpty()) {
            return;
        }
        synchronized (this.attributes) {
            this.attributes.putAll(attributes);
        }
    }`

melekr (approved):

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: Testing getters on app context doesn't make sense. We have tests on the higher level that verify if those information are available

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
konraddysput: Null values will be refused by the metrics system

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
BartoszLitwiniuk: 1. But here you can test logic around Application information cache not only getters
2. Can you point me to place where we have any asserts in tests for these 3 fields?

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
konraddysput: sure - this is also interesting case for the future. In the past we were dropping attributes with nullable value

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+import backtraceio.library.models.Tuple;
+
+public class ReportDataBuilder {
+
+    /**
+     * Divide custom user attributes into primitive and complex attributes and add to this object
+     *
+     * @param attributes client's attributes
+     */
+    public static Tuple<Map<String, String>, Map<String, Object>> getReportAttribues(Map<String, Object> attributes) {
```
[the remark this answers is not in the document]
BartoszLitwiniuk: exactly 👍 
https://github.com/backtrace-labs/backtrace-android/pull/157#issuecomment-2457859396

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
konraddysput: Changing the implementation so this will be more obvious. Thanks for the suggestion

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: my thinking around this:

- initializing cache values once

there is a problem with this - you need to be sure its initialized at least once. You don't know at what point its initialized. Since we have a lot of components that can work together or separated, I have a small problem with this. The best moment is to initialize it during the client initialization, however what if you want to use only database options? 

Cache is necessary to avoid reflection costs.

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
konraddysput: Changed the implementation. I still think nullable values are important for reports (we shouldn't skip them) however for metrics, we cannot send them to API. I extended the interface to make sure we have a flexible option to use them

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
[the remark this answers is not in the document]
konraddysput: fixed

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: Makes sense

konraddysput: Renamed it to ApplicationMetadataCache

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: working on tests - I was fixing the part in the metrics tests that confirms the application name and version are defined. I cannot find them however its hard to manage metrics tests now. I will write dedicated tests for this

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: Moved it to the singleton so we should have optimal solution

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
konraddysput: null check to verify if the first value is defined? It's always defined I think?

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
konraddysput: Reformated file. We should be good now

backtrace-library/src/main/java/backtraceio/library/models/metrics/Event.java
konraddysput: I think we will never run into a situation where due to not safety thread code we can generate an exception. We're managing metrics and the "send" part.

backtrace-library/src/main/java/backtraceio/library/models/json/ReportDataBuilder.java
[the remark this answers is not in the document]
konraddysput: Moved it to the dedicated model

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -6,44 +6,68 @@
...
+     * in a single application scope.
+     *
+     * @param context Application context
+     * @return Application metadata cache
+     */
+    public static ApplicationMetadataCache getInstance(Context context) {
+        if (instance == null) {
+            instance = new ApplicationMetadataCache(context);
+        }
+        return instance;
```
perf2711, line 41: should we worry about thread safety?

backtrace-library/src/main/java/backtraceio/library/models/attributes/ReportDataBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Divide custom user attributes into primitive and complex attributes and add to this object
+     *
+     * @param attributes   client's attributes
+     * @param skipNullable define attributes behavior on nullable value. By default all nullable attributes
+     *                     will be included in the report. For some features like metrics, we don't want to send
+     *                     nullable values, because they can generate invalid behavior/incorrect information.
+     * @return Report data attributes divided into attributes and annotations
+     */
+    public static ReportDataAttributes getReportAttributes(Map<String, Object> attributes, boolean skipNullable) {
```
perf2711, line 28: I think you meant `skipNull`. `Nullable` would be a characteristic of variable, i.e. it can be set to null.

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -6,44 +6,68 @@
...
+    }
```
perf2711, line 42: subsequent calls with different contexts will return the same instance. Is that something we want?

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
```diff
@@ -182,17 +182,24 @@ public void enable(BacktraceMetricsSettings settings, String uniqueEventName) {
             throw new IllegalArgumentException("Unique event name must be defined!");
         }
 
-        this.applicationName = ApplicationHelper.getApplicationName(this.getContext());
-        this.applicationVersion = ApplicationHelper.getApplicationVersion(this.getContext());
+        ApplicationMetadataCache applicationMetadata = ApplicationMetadataCache.getInstance(this.getContext());
+        this.applicationName = applicationMetadata.getApplicationName();
+        this.applicationVersion = applicationMetadata.getApplicationVersion();
+        setStartupUniqueEventName(uniqueEventName);
```
perf2711, line 188: should this be called twice (here and line 191)?

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
```diff
@@ -272,7 +279,7 @@ public boolean addUniqueEvent(String attributeName, Map<String, Object> attribut
             return false;
         }
 
-        Map<String, String> metricsAttributes = ReportDataBuilder.getReportAttribues(attributes).first;
+        Map<String, String> metricsAttributes = ReportDataBuilder.getReportAttributes(attributes, true).getAttributes();
```
perf2711, line 282: it's a bit pointless to collect all attributes and annotations and discard the annotations. Maybe it's worth to create a separate method to get only attributes?

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: We expect to have a single app context

backtrace-library/src/main/java/backtraceio/library/models/attributes/ReportDataBuilder.java
konraddysput: Oh right

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: It shouldn't be a problem in this case, however lets make sure nothing bad will happen. Adding "double checked lock in pattern"

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
konraddysput: this is a bug. Fixed

backtrace-library/src/main/java/backtraceio/library/services/BacktraceMetrics.java
konraddysput: in the metrics context, we won't have too many annotations. We can ofc have a separated method only for attributes, however right now it won't provide too many benefits.

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
perf2711: then maybe throw an error if the context is different?

backtrace-library/src/main/java/backtraceio/library/models/attributes/ReportDataBuilder.java
perf2711: can you also change `nullable` to `null` in the param doc?

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: There is no way you can have two app context. I'm not even sure if this is even possible. The description makes it clear. I don't think it's a good thing. I prefer not to include this

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -0,0 +1,102 @@
...
+    private String packageName;
+
+    /**
+     * Returns current application cache. This instance is a singleton since we can only operate
+     * in a single application scope.
+     *
+     * @param context Application context
+     * @return Application metadata cache
+     */
+    public static ApplicationMetadataCache getInstance(Context context) {
```
BartoszLitwiniuk, line 37: please move this method below constructor

BartoszLitwiniuk: btw maybe we can split this method into `init` and `getInstance` to init something in `get` method

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -0,0 +1,102 @@
...
+                    instance = new ApplicationMetadataCache(context);
+                }
+            }
+        }
+        return instance;
+    }
+
+    private final Context context;
+
+    public ApplicationMetadataCache(Context context) {
```
BartoszLitwiniuk, line 50: I think this constructor should be private otherwise you can do weird things

backtrace-library/src/main/java/backtraceio/library/models/Tuple.java
```diff
@@ -0,0 +1,11 @@
+package backtraceio.library.models;
+
+public class Tuple<T1, T2> {
```
BartoszLitwiniuk, line 3: you are not using this class

backtrace-library/src/main/java/backtraceio/library/models/attributes/ReportDataAttributes.java
```diff
@@ -0,0 +1,27 @@
+package backtraceio.library.models.attributes;
+
+import java.util.HashMap;
+
+public class ReportDataAttributes {
+    private final HashMap<String, String> reportAttributes = new HashMap<>();
```
BartoszLitwiniuk, line 6: use generic `Map` -> Map<String, String> reportAttributes; here and below

backtrace-library/src/main/java/backtraceio/library/models/attributes/ReportDataBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+            if (TypeHelper.isPrimitiveOrPrimitiveWrapperOrString(value.getClass())) {
+                reportDataAttributes.addAttribute(key, value.toString());
+            } else {
+                reportDataAttributes.addAnnotation(key, value);
+            }
+        }
+
+        return reportDataAttributes;
+
```
BartoszLitwiniuk, line 52: remove this line we don't need it

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Retrieves package name from the context.
+     *
+     * @return current package name.
+     */
+    public String getPackageName() {
+        if (!BacktraceStringHelper.isNullOrEmpty(packageName)) {
+            return packageName;
+        }
+        packageName = context.getApplicationContext().getPackageName();
```
BartoszLitwiniuk, line 98: `return context.getApplicationContext().getPackageName();` without var assignment

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -0,0 +1,102 @@
...
+     * @return current application version.
+     */
+    public String getApplicationVersion() {
+        if (!BacktraceStringHelper.isNullOrEmpty(applicationVersion)) {
+            return applicationVersion;
+        }
+        try {
+            PackageInfo info = context.getPackageManager()
+                    .getPackageInfo(context.getPackageName(), 0);
+            applicationVersion = BacktraceStringHelper.isNullOrEmpty(info.versionName) ? String.valueOf(info.versionCode) : info.versionName;
```
BartoszLitwiniuk, line 80: `return BacktraceStringHelper.isNullOrEmpty(info.versionName) ? String.valueOf(info.versionCode) : info.versionName;` you don't have to assign it

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
```diff
@@ -0,0 +1,102 @@
...
+     * Retrieves application name from context. The name will be cached over checks
+     *
+     * @return application name
+     */
+    public String getApplicationName() {
+        if (!BacktraceStringHelper.isNullOrEmpty(applicationName)) {
+            return applicationName;
+        }
+
+        applicationName = context.getApplicationInfo().loadLabel(context.getPackageManager()).toString();
```
BartoszLitwiniuk, line 64: return `context.getApplicationInfo().loadLabel(context.getPackageManager()).toString();` you don't have to assign it

backtrace-library/src/main/java/backtraceio/library/common/DeviceAttributesHelper.java
```diff
@@ -255,33 +262,20 @@ private BatteryState getBatteryState() {
...
-    private String getMaxRamSize() {
-        return Long.toString(getMemoryInformation().totalMem);
-    }
+        String androidId = Settings.Secure.getString(this.context.getContentResolver(),
+                Settings.Secure.ANDROID_ID);
 
-    private String getDeviceFreeRam() {
-        return Long.toString(getMemoryInformation().availMem);
-    }
+        // if the android id is not defined we want to cache at least guid
```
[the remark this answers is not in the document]
konraddysput: one sec - this is the "guid" attribute. This is not report uuid but unique guid

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: how you can cache it?

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: we want to cache the value. Thats the points of this classs

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: We want to cache the value. Thats the point of the class

backtrace-library/src/main/java/backtraceio/library/common/ApplicationMetadataCache.java
konraddysput: re second advice - I would prefer to stay as it is

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceAttributes.java
konraddysput: Added a null checks to be sure we're not adding nullable/empty client attributes.

BartoszLitwiniuk (approved):
lgtm

perf2711 (approved):

backtrace-library/src/main/java/backtraceio/library/common/ApplicationHelper.java
konraddysput: I didn't want to change this code proposed by @BartoszLitwiniuk. I also checked why it was implemented this way - it looks like we might not have access to the BuildConfig when we're working with context. Some info from the Internet:

Advantages of Using BuildConfig
- Simplicity: BuildConfig gives direct access to the version information without needing to call PackageManager or handle exceptions.
- Efficiency: Accessing BuildConfig is faster as it doesn’t require querying the system at runtime.

When to Use PackageManager Instead
The PackageManager approach is mainly useful if:

- You’re trying to retrieve the version information of another app installed on the device.
- You’re working with a context where BuildConfig might not be accessible (e.g., in some library modules).

PR #158, opened 2024-11-19
Nested exception support

konraddysput:
# Why

This pull request captures nested exception via UncaughtExceptionHandler. In the past, the SDK allowed only capturing a single exception, even if the exception had a cause. This pull request converts an exception chain into a report array with additional attribute identifiers, so the user can see the whole nested exception history.

Example:

```
 public void unhandledException(View view) throws IOException {
        try {
            getSaveData();
        } catch (IOException exception) {
            this.throwOriginalExceptionAsInnerException(exception);
        }
    }

    private void throwOriginalExceptionAsInnerException(Exception exception) throws IOException {
        throw new IOException("test", exception);
    }
```
ref: BT-5223


melekr (approved):
lgtm

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -80,4 +84,26 @@ private OnServerResponseEventListener getCallbackToDefaultHandler(final Thread t
...
+        BacktraceReport parent = null;
+
+        List<BacktraceReport> reports = new ArrayList<>();
+        while (throwable != null) {
+            Exception currentException = throwable instanceof Exception ? (Exception) throwable : new UnhandledThrowableWrapper(throwable);
+            BacktraceReport report = new BacktraceReport(currentException, BacktraceExceptionHandler.customAttributes);
+
+            report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
+            report.attributes.put("error.trace", exceptionTrace);
+            report.attributes.put("error.id", report.uuid.toString());
```
perf2711, line 99: why do we need a duplicate uuid?

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -52,9 +55,10 @@ public void uncaughtException(final Thread thread, final Throwable throwable) {
         OnServerResponseEventListener callback = getCallbackToDefaultHandler(thread, throwable);
 
         BacktraceLogger.e(LOG_TAG, "Sending uncaught exception to Backtrace API", throwable);
-        BacktraceReport report = new BacktraceReport(this.getCausedException(throwable), BacktraceExceptionHandler.customAttributes);
-        report.attributes.put(BacktraceAttributeConsts.ErrorType, BacktraceAttributeConsts.UnhandledExceptionAttributeType);
-        this.client.send(report, callback);
+        for (BacktraceReport report :
+                this.transformExceptionIntoReports(throwable)) {
+            this.client.send(report, callback);
+        }
```
perf2711, line 61: Shouldn't this be done in an other place? What if someone sends an exception manually through `Send`?

IMO this should be done when the report is being converted into data acceptable by Backtrace.

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -80,4 +84,26 @@ private OnServerResponseEventListener getCallbackToDefaultHandler(final Thread t
...
+            report.attributes.put("error.parent", parent != null ? parent.uuid.toString() : null);
+            reports.add(report);
+
+            throwable = throwable.getCause();
+            parent = report;
+        }
+
+        return reports;
+    }
 }
```
perf2711, line 109: endline!

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
konraddysput: uuid is not available in the attributes, to make it clear what the parent refers to, I think error.id makes sense here

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
konraddysput: Planning to move it to the Client. We should also consider doing it via database instead at this point if the database is enabled

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -80,4 +84,26 @@ private OnServerResponseEventListener getCallbackToDefaultHandler(final Thread t
             signal.countDown();
         };
     }
+
+    private List<BacktraceReport> transformExceptionIntoReports(Throwable throwable) {
+        final String exceptionTrace = UUID.randomUUID().toString();
+        BacktraceReport parent = null;
```
BartoszLitwiniuk, line 90: You don't need whole BacktraceReport, you need String parentId

konraddysput: The code was moved to another file. We're using string now

backtrace-library/src/main/java/backtraceio/library/models/BacktraceExceptionHandler.java
```diff
@@ -80,4 +84,26 @@ private OnServerResponseEventListener getCallbackToDefaultHandler(final Thread t
...
+
+        List<BacktraceReport> reports = new ArrayList<>();
```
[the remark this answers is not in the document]
konraddysput: The code was moved to another file. It's final now

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,134 @@
...
+        List<String> attachments = sourceReport.attachmentPaths;
+        reports.addAll(this.getSuppressedReports(throwable, attachments, attributes, exceptionTrace, parentId));
+        if (!sendInnerExceptions) {
+            return reports;
+        }
+        throwable = throwable.getCause();
+
+        while (throwable != null) {
+            BacktraceReport report = new BacktraceReport(throwable, attributes);
+            report.attachmentPaths.addAll(attachments);
```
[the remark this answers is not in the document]
konraddysput: talked about this via slack. Changed the approach in favour of UI doing that if necessary

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,134 @@
...
+        List<BacktraceReport> reports = new ArrayList<>();
+        if (!this.sendSuppressedExceptions) {
+            return reports;
+        }
+
+        for (Throwable suppressedException :
+                throwable.getSuppressed()) {
+            BacktraceReport suppressedExceptionReport = new BacktraceReport(suppressedException, attributes);
+            this.extendReportWithNestedExceptionAttributes(suppressedExceptionReport, exceptionTrace, parentId);
+            suppressedExceptionReport.attachmentPaths.addAll(attachments);
```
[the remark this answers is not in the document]
konraddysput: talked about this via slack. Changed the approach in favour of UI doing that if necessary

perf2711 (approved):

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.UUID;
+
+import backtraceio.library.models.json.BacktraceReport;
+
+public class ReportExceptionTransformer {
+
+    /**
+     * Error trace attribute key
+     */
+    public static final String ErrorTraceAttribute = "error.trace";
```
BartoszLitwiniuk, line 15: start field always with a lower case letter

BartoszLitwiniuk, line 15: can be private

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Current error id attribute key
+     */
+    public static final String ErrorIdAttribute = "error.id";
```
BartoszLitwiniuk, line 20: can be private

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Parent id attribute key
+     */
+    public static final String ErrorParentIdAttribute = "error.parent";
```
BartoszLitwiniuk, line 25: can be private

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
```diff
@@ -167,40 +162,53 @@ public BacktraceReport(
...
+        Map<String, Object> reportAttributes = report.attributes != null ? report.attributes :
+                new HashMap<String, Object>();
+        if (attributes == null) {
+            return reportAttributes;
         }
+        reportAttributes.putAll(attributes);
+        return reportAttributes;
+    }
+
+    public String getExceptionClassifier(Throwable exception) {
```
BartoszLitwiniuk, line 201: I think it can be private

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Transforms throwable into an array of Backtrace reports. During the transformation, inner exception and suppressed exceptions
+     * are being converted into Backtrace reports with a proper exception reference.
+     *
+     * @param sourceReport BacktraceReport
+     * @return list of Backtrace reports
+     */
+    public List<BacktraceReport> transformReportWithInnerExceptions(BacktraceReport sourceReport) {
+        final List<BacktraceReport> reports = new ArrayList<BacktraceReport>() {{
```
BartoszLitwiniuk, line 56: check if sourceReport is not null

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Add exception trace attributes to the nested exception
+     *
+     * @param report         Backtrace report
+     * @param exceptionTrace trace id
+     * @param parentId       parent id
+     */
+    private void extendReportWithNestedExceptionAttributes(BacktraceReport report, String exceptionTrace, String parentId) {
```
BartoszLitwiniuk, line 125: I think it can be static because you are not using any object fields

BartoszLitwiniuk:
Are you sure that we should mix changing exception to throwable with adding new feature?

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
```diff
@@ -167,40 +162,53 @@ public BacktraceReport(
...
-            Exception exception,
-            Map<String, Object> attributes,
-            List<String> attachmentPaths) {
-
+    public BacktraceReport(Throwable throwable, Map<String, Object> attributes, List<String> attachmentPaths) {
         this.attributes = CollectionUtils.copyMap(attributes);
         this.attachmentPaths = CollectionUtils.copyList(attachmentPaths);
-        this.exception = this.prepareException(exception);
+        this.originalException = throwable;
+        this.exception = this.prepareException(throwable);
```
BartoszLitwiniuk, line 173: I think this 2 should be combined
```
if (throwable != null) {
    this.exception = this.prepareException(throwable);
    this.classifier = getExceptionClassifier(throwable);
}
```

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
konraddysput: Changed - side note: this is leftover from the previous version

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
konraddysput: I want to use it in other places. It's not a bad to keep it public.

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
konraddysput: I want to use it in other places. It's not bad to keep it public.

backtrace-library/src/main/java/backtraceio/library/services/ReportExceptionTransformer.java
konraddysput: renamed.

backtrace-library/src/main/java/backtraceio/library/models/json/BacktraceReport.java
konraddysput: Adjusted the code to better explain what we do now

BartoszLitwiniuk:
lgtm! Good job @konraddysput

BartoszLitwiniuk (approved):

PR #160, opened 2024-12-21
Tech debt fix -  potential batch null

BartoszLitwiniuk:
Resolve TODO - batch element can be null

### DIFF
--- a/backtrace-library/src/main/java/backtraceio/library/services/BacktraceDatabaseContext.java
+++ b/backtrace-library/src/main/java/backtraceio/library/services/BacktraceDatabaseContext.java
@@ -88,7 +88,7 @@ public class BacktraceDatabaseContext implements DatabaseContext {
         }
 
         for (int i = 0; i < _retryNumber; i++) {
-            this.batchRetry.put(i, new ArrayList<BacktraceDatabaseRecord>());
+            this.batchRetry.put(i, new ArrayList<>());
         }
     }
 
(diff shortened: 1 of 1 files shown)


backtrace-library/src/main/java/backtraceio/library/services/BacktraceDatabaseContext.java
```diff
@@ -372,4 +371,21 @@ private BacktraceDatabaseRecord getRecordFromCache(boolean reverse) {
         }
         return null;
     }
+
+    private void addToFirstBatch(BacktraceDatabaseRecord backtraceDatabaseRecord) {
+        final int firstBatch = 0;
+
+        if (this.batchRetry.isEmpty()) {
```
konraddysput, line 378: how is this even possible if in the custructor, you're making sure at least one batch is being created? L90

BartoszLitwiniuk: it's IDE thing - IntelliJ is not sure if it's initialized because `get` method has annotation `Nullable` but I think it's not that bad to have this extra check which I've introduce if in the future someone by mistake will change this code a bit

BartoszLitwiniuk: + before my change it was doable when someone will pass e.g. _retryNumber = -1

konraddysput (approved):

melekr (approved):

PR #190, opened 2025-10-01
Shadow GSON library 

BartoszLitwiniuk:


backtrace-library/build.gradle
```diff
@@ -115,10 +113,13 @@ mavenPublishing {
...
 
+def gsonJar = project(':gson')
+        .layout.buildDirectory.file("libs/gson-backtrace.jar")
+
 dependencies {
-    implementation 'com.google.code.gson:gson:2.13.1'
-    implementation 'androidx.appcompat:appcompat:1.6.1'
     implementation 'com.squareup:tape:1.2.3'
+    implementation 'androidx.appcompat:appcompat:1.6.1'
+    implementation files(gsonJar)
```
melekr: Dependency duplication risk here ~
Are we sure that both `com.google.code.gson:gson` and the `gson-backtrace` shaded jar aren’t ending up in the AAR?

backtrace-library/build.gradle
```diff
@@ -115,10 +113,13 @@ mavenPublishing {
     }
 }
 
+def gsonJar = project(':gson')
+        .layout.buildDirectory.file("libs/gson-backtrace.jar")
```
melekr: why did we chose this over project dependency?
I'm thinking along the lines of `implementation project(':gson')`

backtrace-library/src/main/java/backtraceio/library/models/BacktraceApiResult.java
```diff
@@ -1,6 +1,6 @@
 package backtraceio.library.models;
 
-import com.google.gson.annotations.SerializedName;
+import backtraceio.gson.annotations.SerializedName;
```
melekr: Since the import switch from `com.google.gson` to `backtraceio.gson` affects all public models, is this considered a breaking change?

example-app/build.gradle
```diff
@@ -72,3 +72,5 @@ dependencies {
     implementation project(':backtrace-library')
     androidTestImplementation project(path: ':coroner-client')
 }
+
+tasks.preBuild.dependsOn(':gson:shadowJar')
```
melekr: What’s the reason for explicitly calling this?
I think `implementation project(':gson')` should handle this automatically ~

backtrace-library/build.gradle
melekr: we might need to look out for transitive dependencies/shaded classes when publishing ~

melekr (changes requested):

example-app/proguard-rules.pro
```diff
@@ -21,5 +21,5 @@
 #-renamesourcefileattribute SourceFile
 
 #These are needed since Proguard breaks some Backtrace libraries
--keep class com.google.gson.**.* { *; }
+-keep class backtraceio.gson.**.* { *; }
```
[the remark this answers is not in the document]
BartoszLitwiniuk: I think we can keep both rules in `consumerProguardFiles` and then everything should be fine

backtrace-library/build.gradle
BartoszLitwiniuk: I wasn't able to configure that in this way that it will use shaded module (after renaming GSON) - it was always using original source of GSON (I think because shading is not done on runtime). We can resolve it easily by extracting whole shaded GSON so separate module which will be published to maven as separate library but it will require extra maintanance, releases, repository etc.

backtrace-library/build.gradle
BartoszLitwiniuk: would be good to verify it with staging repo

melekr (approved):

PR #213, opened 2026-06-15
feature: support more than 64 native attributes

KishanPRao:
## Summary

Removes the 64-entry limit on native (Crashpad) attributes.

Previously, a simple dictionary was used for storing attributes, which limits the total count to 64 entries. They are now stored in Crashpad's global `AnnotationList`.

The attribute count is limited by Crashpad's internal limit, `kMaxNumberOfAnnotations`. `SetAnnotation` will refuse to add keys beyond the limit, providing feedback with a warning.

## Changes

- Removes use of `SimpleStringDictionary`, and replaces it with `AnnotationList`. This data is dynamically allocated.
- Native crash reports omit attributes with an empty-string value (this was allowed previously). Managed/JVM reports are unaffected, documented in`addAttribute`.

ref: BT-6868

### DIFF
--- a/backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
+++ b/backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
@@ -61,12 +61,25 @@ namespace {
     }
 
     // Stores attributes as crashpad StringAnnotations, replacing the 64-entry
-    // simple_annotations dictionary. The snapshot reader reads at most
-    // crashpad::kMaxNumberOfAnnotations per module, so SetAnnotation drops new
-    // keys past that. Annotations and name buffers are allocated once and never
+    // simple_annotations dictionary. The crashpad snapshot reader reads at most
+    // crashpad::kMaxNumberOfAnnotations annotation objects per module, so
+    // SetAnnotation drops new user keys past that limit (internal keys keep
+    // reserved slots). Annotations and name buffers are allocated once and never
     // freed (the global list references them for the process lifetime).
     constexpr crashpad::Annotation::ValueSizeType kAnnotationValueMaxSize = 4096;
     using DynamicAnnotation = crashpad::StringAnnotation<kAnnotationValueMaxSize>;
+
+    // Reserve slots for internal, per-report annotations so they are never
+    // crowded out by user attributes near the read limit.
+    constexpr size_t kReservedAnnotationSlots = 2; // error.message, _mod_faulting_tid
+
+    bool IsValidAnnotationKey(const char* rawKey) {
+        return rawKey != nullptr && rawKey[0] != '\0';
+    }
+
+    bool IsReservedAnnotationKey(const std::string& key) {
+        return key == "error.message" || key == "_mod_faulting_tid";
+    }
 
     std::unordered_map<std::string, DynamicAnnotation*> g_annotations;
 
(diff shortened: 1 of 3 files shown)


melekr (commented):
Nice change @KishanPRao, moving dynamic attributes off `SimpleStringDictionary` is the right direction and should remove the 64 entry limit.

I'm adding a few small correctness/edge-case notes below, mostly around making the new behaviour explicit and covered by tests, should be straightforward to address.

melekr:
Crashpad still caps annotation list at **[kMaxNumberOfAnnotations](https://github.com/backtrace-labs/crashpad/commit/9c58b668ff3372448b1c2ef6e1f3c34930825862)**, currently `200` on the main fork but I think we updated it a while ago to `400`, so this is not strictly `unbounded`. 

Could we update the PR wording/comments to something like “supports more than 64 dynamic attributes” or “supports dynamically allocated attributes up to Crashpad’s annotation list read limit”?

melekr:
> Crashpad still caps annotation list at **[kMaxNumberOfAnnotations](https://github.com/backtrace-labs/crashpad/commit/9c58b668ff3372448b1c2ef6e1f3c34930825862)**, currently `200` on the main fork but I think we updated it a while ago to `400`, so this is not strictly `unbounded`.
> 
> Could we update the PR wording/comments to something like “supports more than 64 dynamic attributes” or “supports dynamically allocated attributes up to Crashpad’s annotation list read limit”?

a regression test proving 65+ dynamic attributes make it into the native report/upload can be useful here. If we want a runtime guard to make the behaviour less surprising, something along these lines could work:

```
// Keep this aligned with Crashpad's snapshot/snapshot_constants.h kMaxNumberOfAnnotations. 
// Crashpad currently reads at most 400 annotation objects from a module.
constexpr size_t kMaxCrashpadAnnotationObjectsRead = 400;

bool WouldExceedCrashpadAnnotationReadLimit(const std::string& key) {
    return g_annotations.find(key) == g_annotations.end()
            && g_annotations.size() >= kMaxCrashpadAnnotationObjectsRead;
}
```

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
```diff
@@ -56,6 +59,50 @@ namespace {
...
+            return it->second;
+        }
+        char* nameCopy = new char[key.size() + 1];
+        std::memcpy(nameCopy, key.c_str(), key.size() + 1);
+        DynamicAnnotation* annotation = new DynamicAnnotation(nameCopy);
+        g_annotations.emplace(key, annotation);
+        return annotation;
+    }
+
+    void SetAnnotation(const char* rawKey, const char* rawValue) {
```
melekr, line 99: One compatibility edge case: empty string values may no longer be represented. 

With `StringAnnotation`, `Set(base::StringPiece(""))` results in size 0, which Crashpad treats as cleared/not included.  The previous `SimpleStringDictionary` considered an entry active based on the key, so an empty value could still be present.

Could we either document that native attributes with empty string values are skipped, or preserve the old behaviour? 
a test would be helpful either way: `client.addAttribute("empty.attribute", "");`

If preserving empty values is important, we may need a special path for empty values, since `AnnotationList` cannot distinguish “set to empty string” from “cleared” using size `0`.

KishanPRao:
> > Crashpad still caps annotation list at **[kMaxNumberOfAnnotations](https://github.com/backtrace-labs/crashpad/commit/9c58b668ff3372448b1c2ef6e1f3c34930825862)**, currently `200` on the main fork but I think we updated it a while ago to `400`, so this is not strictly `unbounded`.
> > Could we update the PR wording/comments to something like “supports more than 64 dynamic attributes” or “supports dynamically allocated attributes up to Crashpad’s annotation list read limit”?
> 
> a regression test proving 65+ dynamic attributes make it into the native report/upload can be useful here. If we want a runtime guard to make the behaviour less surprising, something along these lines could work:
> 
> ```
> // Keep this aligned with Crashpad's snapshot/snapshot_constants.h kMaxNumberOfAnnotations. 
> // Crashpad currently reads at most 400 annotation objects from a module.
> constexpr size_t kMaxCrashpadAnnotationObjectsRead = 400;
> 
> bool WouldExceedCrashpadAnnotationReadLimit(const std::string& key) {
>     return g_annotations.find(key) == g_annotations.end()
>             && g_annotations.size() >= kMaxCrashpadAnnotationObjectsRead;
> }
> ```

Since we'd need to add an entire gtest suite, for native testing, I'd handle this separately, at a later point.

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
KishanPRao: Good point!
For now, I've added this into the documentation, so the behavior is clear:
https://github.com/backtrace-labs/backtrace-android/pull/213/changes#diff-c09d2e0ffa8d425c75de730925f463d041c995ae89cd6ec53d586e90b5b710abR360

melekr (approved):

backtrace-library/src/main/cpp/backends/crashpad-backend.cpp
KishanPRao: Also added doc here, for completeness:

https://github.com/backtrace-labs/backtrace-android/pull/213/changes#diff-b41a4ed90de46b65c07d7eac47dc7c28ecc3e03c4c9e39b4c8ec5421d7f6741cR26

PR #216, opened 2026-07-21
chore: stabilize and speed up Android CI

KishanPRao:
## Summary

Fixes a flaky test on old Sauce real devices and trims CI time (~18m to ~13m in typical runs).

## Changes

* `ExampleInstrumentedTest.java`: re-assert test state before starting.
* `test.yml`: retry tests on `saucectl` to account for device flakiness, run device suites in parallel, move region to `eu-central-1`, cache Gradle deps, add a build profile artifact, checkout the repo during test.
* `.sauce/config.yml`: define the Sauce device suites, real devices + emulators.
* `gradle.properties`: enable parallel build, raise heap.

## Note

The `spotlessCheck` step is removed for CI builds, since it is redundant.

ref: BT-7288


melekr (approved):
awesome 🚀

PR #217, opened 2026-08-05
chore(deps): upgrade Android Gradle Plugin to 8.13

BartoszLitwiniuk:
Upgraded the Android Gradle Plugin from `8.11.1` to `8.13.2` and updated the related Gradle configuration. Also replaced the deprecated `proguard-android.txt` with `proguard-android-optimize.txt` and cleaned up the duplicated `buildTypes` block.

PR #219, opened 2026-08-11
chore(deps): upgrade to AGP 9.3.1 and Gradle 9.5

BartoszLitwiniuk:
Should be merged after https://github.com/backtrace-labs/backtrace-android/pull/217

# Summary

Upgrades the build toolchain to Android Gradle Plugin 9.3.2 (from 8.11.1) and Gradle 9.5.0 (from 8.14.3), migrating the build scripts to the DSL APIs required by AGP 9.

## Changes

**Toolchain**
- AGP `8.11.1` → `9.3.2` (root `build.gradle` classpath and `settings.gradle` plugin declarations)
- Gradle wrapper `8.14.3` → `9.5.0`
- Enable `org.gradle.tooling.parallel`

**DSL migration (required by AGP 9)**
- Convert `android {}` block assignments to property syntax (`namespace =`, `minSdk =`, `versionCode =`, …)
- Move `compileSdk` out of `defaultConfig` to the `android {}` block
- Replace removed `lintOptions {}` with `lint {}`, and declare `targetSdk` under `lint {}` / `testOptions {}` in the library module (`targetSdk` is no longer valid in a library's `defaultConfig`)
- Replace `testCoverageEnabled` with `enableUnitTestCoverage` / `enableAndroidTestCoverage`
- Fix a pre-existing nesting bug in `backtrace-library/build.gradle` where `buildTypes` was declared inside itself, so the `debug` coverage config never applied
- Flatten the redundant nested `android {}` / `defaultConfig {}` blocks in `example-app/build.gradle`

**Publishing**
- Capture `androidGitVersion.name()` into `libVersion` and use it for the Maven coordinates, since `android.defaultConfig.versionName` is no longer readable at configuration time for library modules
- Update Java in publishing pipeline to v21

**ProGuard**
- Switch the example app to `proguard-android-optimize.txt` to silence a build warning
- Library release build now uses `proguard-android-optimize.txt` instead of `proguard-android.txt`

**CI pipeline**
- Update Java in CI pipeline to v21


backtrace-library/build.gradle
```diff
@@ -10,17 +10,16 @@ androidGitVersion {
...
+    compileSdk = 36
     defaultConfig {
-        minSdkVersion 21
-        compileSdk 36
-        targetSdkVersion 36
-        ndkVersion '27.3.13750724'
-        versionName androidGitVersion.name()
+        minSdk = 21
+        ndkVersion = '27.3.13750724'
         buildConfigField("String", "VERSION_NAME", "\"${versionName}\"")
```
melekr: libVersion is now used for Maven coordinates, but removing defaultConfig.versionName leaves the existing `buildConfigField(... "${versionName}")` without a value;
- `backtrace-library:generateDebugBuildConfig` emits BuildConfig.VERSION_NAME = "null". 
- BacktraceBase.version then propagates that into each report’s agentVersion and backtrace.version.

Can we use the captured libVersion for the BuildConfig field as well, and (bonus) add assertion that rejects both null and the literal "null"?

backtrace-library/build.gradle
melekr, line 1: according to [AGP Release notes](https://developer.android.com/build/releases/agp-9-0-0-release-notes) Behavior changes;

> Android Gradle plugin 9.0 by default requires consumers of a library to use the same or higher compile SDK version.
> Use the same or higher compile SDK when consuming a library. If this is not possible, or you want to give consumers of a library you publish more time to switch, set [AarMetadata.minCompileSdk](https://developer.android.com/reference/tools/gradle-api/9.0/com/android/build/api/dsl/AarMetadata#minCompileSdk()) explicitly.

do we need to set `aarMetadata.minCompileSdk` explicitly to the verified compatibility floor?

melekr, line 1: The tag-publishing workflow still selects JDK 17.

melekr, line 1: AGP 9 enables built-in Kotlin automatically, we might need to set `enableKotlin = false `

BartoszLitwiniuk: not setting this won't theoretically cause much harm because we don't have any .kt file anyway, but it might increase the library size a bit, so I'll disable it

backtrace-library/build.gradle
BartoszLitwiniuk: Thanks, It's really valid point, I've overlooked it so good that you've found it. I've fixed it and added var validation.

backtrace-library/build.gradle
BartoszLitwiniuk: Updated to v21

melekr (approved):

backtrace-library/build.gradle
BartoszLitwiniuk: I'm setting it to version 36 to give customers more time to transition in [next PR](https://github.com/backtrace-labs/backtrace-android/pull/220/changes#r3856880003). 

In this PR we don't have to do that because we're not upgrading the API version so it will be still API 36.

KishanPRao: Just to be clear, are we setting `enableKotlin = false`?
I see that the comment has been resolved without the change.

BartoszLitwiniuk: good point, indeed, I resolved the comment without doing a push of my changes. Thanks

PR #223, opened 2026-08-17
Un-pin GitHub Actions

melekr:
Un-pin GitHub Actions for Testing.

> [GitHub Actions](https://docs.github.com/en/actions) are typically referenced by version tags (e.g., actions/checkout@v5), but version tags can be moved, force-pushed, or accidentally updated. For better reproducibility and security, you can pin actions to specific version SHAs (immutable references that never change).

### Why Pin to SHAs?

> Version tags are convenient, but they’re mutable. A maintainer can re-release a tag, move it to a different commit, or even force-push to it. This means your workflow could run different code tomorrow than it did today without you changing anything.
> Here’s the key insight: a version tag like v5.0.1 is just a pointer to a specific commit. When you pin to the commit SHA that tag currently points to, you’re locking in that exact version. If the tag ever gets moved or re-released to point to a different commit, your workflow stays pinned to the original commit. You’re insulated from any future changes to that tag.
> Commit SHAs are different. A 40-character SHA like 93cb6efe18208431cddfb8368fd83d5badbf9bfd points to one specific commit, forever. It can never change, never be re-released, never be overwritten.
> For production workflows and security-sensitive operations, this immutability is invaluable.

refs: 
- https://cloudnativeengineer.substack.com/p/github-actions-reproducibility-security
- https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,20 +53,20 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Checkout submodules
-        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v5
+        uses: actions/checkout@v5
         with:
           fetch-depth: 2
           submodules: recursive
           persist-credentials: false
 
       - name: set up JDK 17
-        uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v5
+        uses: actions/setup-java@v5
         with:
           java-version: "17"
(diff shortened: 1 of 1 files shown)


melekr:
FYI @BartoszLitwiniuk @KishanPRao.

just to show the reasoning behind SHA pinning vs Tags. We'll close this PR / won't merge.

